PR #3, opened 2026-07-07
fix(config): require user-level provider registry

3aKHP:
## Summary

- require the user-level `providers.yaml` instead of falling back to legacy single-key provider env config
- load provider secrets from the `.env` file beside `providers.yaml`, with process env as fallback only
- report provider `.env` status in `doctor` and update docs/tests for the migration path

## Test Plan

- [x] `bun run typecheck`
- [x] `bun test`
- [x] `bun run doctor`

## Notes / Follow-ups

- Migrated the local legacy project-root `.env` to `C:\Users\asus\AppData\Roaming\prism-vesicle\.env`; secret values were not printed.


3aKHP:
Addressed the automated review in `697142a`:

- Fixed the medium TUI issue: `/use`, `/model`, and session resume now call `inspectProviderConfig(selection)` so API-key availability is resolved through the user-level provider `.env`.
- Improved the missing explicit `VESICLE_PROVIDERS_FILE` error message.
- Added a clearer bare `export KEY` parse error for provider `.env` files.
- Cleaned up provider-config temp dirs in config and agent-loop tests.
- Added regression coverage for the explicit missing file case, provider `.env` precedence, and bare export errors.

Verification:

- `bun run typecheck`
- `bun test` (64 pass)
- `bun run doctor`

PR #4, opened 2026-07-07
refactor(provider): split OpenAI chat backend

3aKHP:
## Summary

- split the OpenAI-compatible backend into request shaping, response parsing, streaming, wire types, and structured provider errors
- keep `OpenAIChatCompatibleAdapter` as the orchestration boundary while preserving the public provider entrypoint
- add provider backend tests for request body shape, structured errors, SSE parsing, and provider-layer import boundaries

## Test Plan

- [x] `bun run typecheck`
- [x] `bun test`
- [x] `bun run doctor`

## Notes / Follow-ups

- This intentionally does not add new provider protocols yet; it hardens the existing OpenAI-compatible path as the template for later adapters.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -51,6 +51,10 @@ project follows Semantic Versioning once releases begin.
   provider-stream error, retries without OpenAI-specific `stream_options` for
   stricter compatible providers, and preserves the final assistant turn in the
   in-memory conversation history.
+- OpenAI-compatible reasoning responses now preserve provider
+  `reasoning_content` through non-streaming, streaming, tool-loop follow-up
+  requests, and session resume so reasoning models can use tools without losing
+  required thinking context.
 - TUI provider/model switches now resolve API-key availability through the
   user-level provider `.env`, so `/use`, `/model`, and session resume do not
   incorrectly show "API key: missing" when the selected key is stored there.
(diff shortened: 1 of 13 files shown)


3aKHP:
Addressed the automated review in `1e6d34c`:

- Wrapped `reader.releaseLock()` so an auto-released errored stream does not mask the original stream failure.
- Changed streamed function names to direct assignment instead of concatenation.
- Added backend coverage for `readChatCompletionStream` content happy path, streamed tool-call accumulation, and the adapter `stream()` JSON content-type fallback.

Verification:

- `bun run typecheck`
- `bun test` (73 pass)
- `bun run doctor`

3aKHP:
Added a follow-up fix in `48bccd1` for reasoning-model tool loops:

- Added `reasoningContent` to the provider-neutral message/response IR.
- Parsed OpenAI-compatible `reasoning_content` from non-stream responses and streamed `delta.reasoning_content`.
- Serialized assistant `reasoningContent` back to Chat Completions as `reasoning_content`, including assistant tool-call messages.
- Persisted reasoning content in session metadata and restored it on session resume.
- Added regression coverage for DeepSeek-style reasoning tool loops: the second provider request after a tool result now includes the prior assistant `reasoning_content`.

Verification:

- `bun run typecheck`
- `bun test` (74 pass)
- `bun run doctor`

3aKHP:
Addressed the latest automated review in `ed58b26`:

- Threaded providerId into malformed response and stream ProviderError paths.
- Kept readSseData internal and covered CRLF/multiline SSE behavior through readChatCompletionStream.
- Added explicit coverage for latest-value streamed tool-call function names.
- Resolved provider boundary paths relative to the test file instead of process.cwd().
- Documented the streamed tool-call name handling fix in CHANGELOG.

Verification:

- `bun run typecheck`
- `bun test` (75 pass)
- `bun run doctor`

PR #5, opened 2026-07-07
feat(provider): add runtime thinking tier control

3aKHP:
## Summary

- Add a normalized `ReasoningTier` surface: `off`, `low`, `midium`, `high`, `xhigh`, `max`.
- Add TUI `/think <tier>` runtime control for subsequent turns, with session persistence and resume restoration.
- Thread generation controls through `runPrompt`/`resolveGate` into provider requests.
- Map OpenAI-compatible request controls so `off` disables thinking, `low`/`midium`/`high` use high effort, and `xhigh`/`max` use max effort.
- Keep unset sessions provider-default: no thinking control fields are sent until the user explicitly chooses a tier.

## Verification

- `bun run typecheck`
- `bun test` (78 pass)
- `bun run doctor`

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -35,8 +35,10 @@ Read asset-specific docs when touching assets:
 If a task touches architecture, provider behavior, TUI interaction, session
 semantics, or command UX, look for existing implementation patterns before
 inventing a new one. The ignored local `dev/docs/` area may contain private
-reference-project notes and local paths; use those notes when available, but do
-not copy local absolute paths or private machine details into public docs.
+reference-project notes and local paths. When present, start with
+`dev/docs/REFERENCE_PROJECTS.md` to find the local reference project locations;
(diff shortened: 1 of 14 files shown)


3aKHP:
Follow-up in `bbf1d36` after checking the reference-project pattern, especially ClaudeCode `/effort`:

- Added `/think auto` plus `unset`/`default` aliases to clear an explicit thinking tier and return to provider/model defaults.
- Persist clearing as `reasoningTier: null` so session resume restores the provider-default state correctly.
- Updated README, STATUS, CHANGELOG, and STYLE for the reset behavior.
- Updated AGENTS.md and CLAUDE.md to prefer documented reference-project patterns before reinventing agent/runtime behavior, while keeping local absolute paths confined to ignored private `dev/docs` notes.

Verification:

- `bun run typecheck`
- `bun test` (79 pass)
- `bun run doctor`

PR #6, opened 2026-07-07
feat(tui): show collapsible reasoning content

3aKHP:
## Summary

- surface provider `reasoning_content` through agent-loop streaming events and final assistant response events
- render reasoning as a separate TUI thinking block with `/reasoning hidden|collapsed|expanded`, default collapsed bounded preview, and resumed-session support
- document reasoning visibility behavior and the RikkaHub-style separate reasoning block pattern

## Test Plan

- [x] `bun run typecheck`
- [x] `bun test tests\agent-loop.test.ts tests\provider-backend.test.ts tests\tui.test.tsx`
- [x] `bun test`
- [x] `bun run doctor`
- [x] manual TUI smoke by user

## Notes / Follow-ups

- This keeps reasoning metadata separate from assistant prose; native Anthropic/Gemini/OpenAI Responses thinking surfaces remain deferred with their adapters.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,9 @@ project follows Semantic Versioning once releases begin.
   control, plus `/think auto`/`unset` to return to provider defaults. Selected
   tiers are passed through the agent loop, persisted in session metadata, and
   restored on resume.
+- TUI `/reasoning hidden|collapsed|expanded` command and independent thinking
+  blocks for provider `reasoning_content`, including live streamed reasoning
+  display and bounded collapsed/expanded tail views.
 
 ### Changed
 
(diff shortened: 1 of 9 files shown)


3aKHP:
Addressed the Bot Review follow-ups in e408304:

- persisted `/reasoning` display mode in session metadata and restored it via `loadSessionSnapshot()`
- replaced the resumed-message cast with an explicit `ResumedMessage[] -> VesicleMessage[]` mapper
- documented `/reasoning` aliases in help/status usage text
- added a session regression test for restoring the latest reasoning display mode

Verification refreshed:
- `bun run typecheck`
- `bun test tests\session.test.ts tests\agent-loop.test.ts tests\tui.test.tsx`
- `bun test`
- `bun run doctor`

PR #7, opened 2026-07-07
feat(config): support model generation defaults

3aKHP:
## Summary

- allow `providers.yaml` models to be either string entries or object entries with `id`, `generation`, and `capabilities`
- merge selected model generation defaults in the agent loop while keeping high-frequency thinking control as the TUI override
- remove the OpenAI Chat adapter's hardcoded temperature default and show model defaults/capabilities in `/models`
- move the provider secret template beside the provider registry template under `docs/examples/`

## Test Plan

- [x] `bun run typecheck`
- [x] `bun test`
- [x] `bun run doctor`

## Notes / Follow-ups

- Existing user-level string model lists remain compatible; the current local `%APPDATA%\prism-vesicle\providers.yaml` shape does not require migration.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -28,6 +28,7 @@ Read asset-specific docs when touching assets:
 
 - `assets/README.md`: Prism asset source and adaptation notes.
 - `docs/examples/providers.yaml`: canonical provider registry shape.
+- `docs/examples/provider.env.example`: user-level provider secret file shape.
 - `dev/docs/working/PLAN-Prism-Vesicle-architecture.md`: older architecture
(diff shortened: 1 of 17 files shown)


3aKHP:
Addressed the Bot Review follow-ups in b2dc8cf:

- only entries starting with `id:` are parsed as object model entries, so colon-bearing string model IDs such as `anthropic:claude-3-opus` remain valid
- generation merging now copies only defined values, so undefined overrides do not erase configured model defaults
- added regression coverage for both cases

Verification refreshed:
- `bun run typecheck`
- `bun test tests\config.test.ts tests\agent-loop.test.ts`
- `bun test`
- `bun run doctor`

PR #8, opened 2026-07-07
feat(providers): preserve thinking blocks

3aKHP:
## Summary

- introduce provider-neutral `thinkingBlocks` on Vesicle messages and responses while keeping `reasoningContent` as an OpenAI-compatible bridge
- map OpenAI Chat `reasoning_content` into replayable thinking blocks and back into follow-up Chat Completions requests
- persist/reload thinking blocks through the agent loop, session JSONL metadata, and resumed TUI transcript display
- document the protocol-state boundary as groundwork for native Anthropic Messages and Gemini adapters

## Test Plan

- [x] `bun run typecheck`
- [x] `bun test tests\provider-backend.test.ts tests\agent-loop.test.ts tests\session.test.ts tests\tui.test.tsx`
- [x] `bun test`
- [x] `bun run doctor`

## Notes / Follow-ups

- This PR does not add a new protocol adapter; it prepares the internal state shape so Anthropic/Gemini thinking blocks can be preserved without flattening them into assistant prose.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,10 @@ project follows Semantic Versioning once releases begin.
 
 ### Changed
 
+- Reasoning preservation now uses provider-neutral thinking blocks internally,
+  while keeping OpenAI-compatible `reasoningContent` as a compatibility bridge.
+  Agent-loop events, session replay, and TUI display can now carry provider
+  thinking state without flattening it into assistant prose.
 - Expanded `AGENTS.md` and `CLAUDE.md` into full AI-collaborator entry points
   that link the repo's workflow, style, status, contribution, provider config,
   verification, and documentation-sweep rules.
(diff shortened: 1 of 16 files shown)


3aKHP:
Addressed the Bot Review follow-ups in 044f7e4:

- `reasoningContentFromThinkingBlocks()` now concatenates multiple reasoning blocks instead of returning only the first one
- session replay now restores only known thinking block shapes with the expected payload field types
- `displayTextFromThinkingBlocks()` now keeps explicit `continue` flow for `thought_summary`
- added helper and session regression tests

Verification refreshed:
- `bun run typecheck`
- `bun test tests\thinking.test.ts tests\session.test.ts tests\provider-backend.test.ts tests\agent-loop.test.ts`
- `bun test`
- `bun run doctor`

PR #11, opened 2026-07-07
feat(providers): add Gemini generateContent protocol

3aKHP:
## Summary

- Add Gemini generateContent provider support for non-streaming text, function calls, function responses, thinking controls, and thoughtSignature replay.
- Add streamGenerateContent SSE support with content, reasoning, function-call deltas, and final response reconstruction.
- Update provider config examples and docs for the new gemini-generate-content protocol and x-goog-api-key auth method.

## Test Plan

- [x] bun run typecheck
- [x] bun test tests\\gemini-provider.test.ts tests\\config.test.ts
- [x] bun test
- [x] bun run doctor

## Notes / Follow-ups

- Official Gemini generateContent uses part-level thoughtSignature state, so the adapter preserves replayable gemini_part blocks instead of flattening them into assistant prose.


3aKHP:
Addressed Bot Review follow-up in a0b67d2:

- Fixed Gemini stream reconstruction for cumulative GenerateContentResponse chunks by processing only unseen prefix-matched parts while still supporting delta-style chunks.
- Added a cumulative-parts regression test.
- Extracted shared SSE parsing to src/providers/shared/sse.ts and switched Anthropic + Gemini streaming paths to use it.
- Documented Gemini's atomic functionCall args behavior at the stream event emission site.

Verification after the fix:
- bun run typecheck
- bun test tests\\gemini-provider.test.ts tests\\anthropic-provider.test.ts tests\\provider-backend.test.ts tests\\provider-stream.test.ts
- bun test

PR #13, opened 2026-07-07
feat(tools): record file operation events

3aKHP:
## Summary

- Add structured fileEvent metadata to successful filesystem tool results.
- Persist fileEvent metadata on agent-loop tool records in session JSONL and emit it through tool_result activity events.
- Update docs and tests for the file-operation ledger contract.

## Test Plan

- [x] bun run typecheck
- [x] bun test tests\\file-tools.test.ts tests\\agent-loop.test.ts tests\\session.test.ts
- [x] bun test
- [x] bun run doctor

## Notes / Follow-ups

- This PR records the backend ledger data only; TUI artifact/history views can consume fileEvent metadata in a later UI-focused pass.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,9 @@ project follows Semantic Versioning once releases begin.
   `create_file`, `replace_in_file`, `append_file`, `delete_file`, `copy_file`,
   and `move_file`, all behind the existing project-relative path guards and
   artifact-root write boundaries.
+- File tool operation ledger: successful filesystem tools now emit structured
+  `fileEvent` metadata through agent-loop activity events and session JSONL
+  tool records.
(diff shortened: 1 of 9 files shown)


3aKHP:
Addressed CR in be123fc:

- Clarified FileToolEvent metric semantics: bytes is the resulting/observed/deleted file size, append_file also reports deltaBytes, list_files reports entryCount, grep_files reports matches.
- Reused assertFile() stat results for copy/move/delete and recorded deleted file size.
- Added operation-specific metadata assertions for create/read/replace/append/list/copy/move/delete plus existing stat/grep checks.
- Documented field semantics in STYLE.md.

Verification after the fix:
- bun run typecheck
- bun test tests\\file-tools.test.ts tests\\agent-loop.test.ts tests\\session.test.ts
- bun test
- bun run doctor

PR #15, opened 2026-07-12
feat(agents): add asynchronous subagent runtime

3aKHP:
## Summary

- add profile-driven foreground/background SubAgents with parallel execution, durable parent inbox delivery, cancellation, messaging, explicit waits, crash reconciliation, and parent/child write coordination
- add host-only run IDs plus short parent-scoped handles, legacy ID compatibility, built-in/custom Agent Profiles, and model-visible control tools
- add first-class AgentCard lifecycle UI, header/sidebar summaries, parked cross-session visibility, `/agents` detail/stop/retry commands, and stable parent STATUS ownership
- serialize same-session appends, attribute child mutations to parent rewind checkpoints, suppress cancelled-agent wakeups, and make recovery idempotently repair missing foreground results/background inbox entries
- document the runtime contract and synchronize README, STATUS, changelog, prompt assets, and bilingual user guidance

## Test Plan

- [x] `bun run typecheck`
- [x] `bun test` (398 passed, 0 failed)
- [x] `bun run doctor`
- [x] `bun run build:assets`
- [x] `bun run pack:check` (165 package files)
- [x] `git diff --check`
- [x] focused 80-column AgentCard, async/parallel, cancellation, recovery, session serialization, checkpoint, completion, and legacy compatibility tests

## Independent CR

- completed clean-context independent review
- addressed all Blocking and Should-fix findings before push, including foreground/background crash windows, session append races, restoration/delivery ordering, parent checkpoint attribution, terminal control races, and truthful recovery result status
- final review reported no remaining Blocking findings

## Notes / Follow-ups

- real-provider E2E remains opt-in and was skipped because `BUN_E2E_REAL_PROVIDER` was not set
- SubAgent recursion and in-flight provider-request replay after process exit remain intentionally out of scope and are documented in `STATUS.md`


3aKHP:
## Deferred review note

The session-append failure policy raised in review is intentionally deferred from this PR.

The append call that fails already rejects to its caller. This PR keeps later queued appends eligible to proceed instead of permanently poisoning the entire session after one transient filesystem error. Whether Vesicle should instead mark a session degraded and reject every subsequent append is a broader durability and product-policy decision that needs explicit recovery semantics and fault-injection coverage.

No change is being made to that policy in PR #15.

PR #18, opened 2026-07-13
feat(shell): add managed background execution

3aKHP:
## Summary

- add `shell_exec.runInBackground` with durable `shell-N` task state, bounded output, completion delivery, and explicit `shell_output` / `shell_stop` controls
- harden Process Runtime cleanup, timeout handling, permission recovery, and effective-tool-surface validation
- surface foreground/background shell progress in tool cards, the header, and a fixed-height Workspace sidebar summary
- update the prompt contract, architecture/status docs, changelog, README, and bilingual shell manual

## Security and recovery

- background commands retain host-user authority and remain permission-controlled
- YOLO cannot execute a provider-hallucinated tool outside the effective Engine surface
- background state is persisted under ignored `.vesicle/processes/`; stale running records recover as interrupted and are never replayed
- shell-tainted rewind warnings and indeterminate-process recovery remain fail-closed

## Verification

- `bun run typecheck`
- `bun test` — 454 passing
- `bun run doctor`
- `git diff --check`
- focused 80/100-column TUI regression with three concurrent background shells

## Notes

This is a high-risk model-visible tool/runtime change and targets `develop` through the normal PR path. No release tag or `main` update is included.


3aKHP:
Review follow-up in 593758e:

- cleaned up both timeout and AbortSignal observers after every `ProcessManager.wait` race outcome
- split the fail-closed parent-session requirement from the ownership comparison and added explicit missing-session regression coverage for `shell_output` and `shell_stop`
- unified process-card duration formatting with the shared tool renderer

The completion observer remains bounded by the approved process plan timeout (`shell_exec.timeoutMs`, maximum 600000 ms), so it cannot retain the session closure indefinitely. Full local verification and all three CI jobs pass on the updated head.

PR #20, opened 2026-07-13
refactor(agent-loop): split turn orchestration

3aKHP:
## Summary

- reduce the agent-loop public entrypoint to a stable facade and move turn bootstrap, provider rounds, planning, execution, interaction pauses, continuation paths, and finalization into domain modules
- centralize tool-result message, session, event, metadata, and background-process recording while preserving permission, checkpoint, SubAgent concurrency, and continuation ordering
- replace line-count thresholds in the architecture style guide with semantic cohesion, coupling, testability, and change-safety criteria

## Test Plan

- [x] bun test tests/agent-loop-round.test.ts tests/agent-loop.test.ts
- [x] bun run typecheck
- [x] bun test
- [x] bun run doctor
- [x] bun run build:exe
- [x] git diff --check

## Notes / Follow-ups

- This is an internal refactor; provider wire formats, session JSONL shape, tool metadata, permission semantics, and public agent-loop exports are unchanged.
- The real-provider E2E remains opt-in and was not run; the deterministic gate E2E path passed in the full suite.


3aKHP:
## Independent CR

Blocking: none.

Should-fix: none.

Nits: none.

Verified:
- public agent-loop exports remain stable through the run.ts facade
- tool-result message, session metadata, background tracking, and event ordering remain intact
- permission resume preserves durable request validation, unresolved pairing, deferred Agent batching, capability drift, and exact shell plan hashes
- already-started SubAgents still drain and persist after sibling host-tool failures
- interaction redirects and pauses preserve provider tool-call pairing
- background shell notification behavior remains covered

Residual risk: the opt-in real-provider E2E was not run. Deterministic local coverage passed with 462 tests, and PR CI passed typecheck/tests plus Linux ELF and Windows PE smoke.

PR #21, opened 2026-07-14
feat(harness): add managed pack foundation

3aKHP:
## Summary

- add strict `prism-harness-pack/v1` parsing, inventory/hash verification, Profile/Prompt and embedded Adapter/Rule binding checks
- add fail-closed host capability reporting and immutable staging-based installation for already-extracted packs
- extract shared child-Agent tool-scope validation so Harness-provided Agent Profiles cannot expose recursive or interactive controls
- document the managed complete-baseline boundary and keep `/permissions` as the sole tool-call authorization layer

## Test Plan

- [x] `bun run typecheck`
- [x] `bun test` (468 pass)
- [x] `bun run doctor`
- [x] `bun run build:assets`
- [x] `bun run pack:check` (231 files)
- [x] `bun run build:exe` (Windows PE + Linux ELF)
- [x] `git diff --check`
- [x] verify Neural-Narratology prerelease `harness-20260713-1` against the foundation
- [x] independent CR: 0 Blocking / 0 Should-fix / 0 Nits

## Notes / Follow-ups

- Runtime activation, project lock/selection, ZIP extraction, rollback, and online release discovery remain deferred.
- The current prerelease remains intentionally incompatible because Vesicle does not yet implement `prism-agent/delegation@1` or `quality-guard/anti-ai-flavor@1`.
- A selected managed Harness must later be one complete baseline; missing pack files must not fall through to bundled V9 except for explicit `externalHostAssets`.


3aKHP:
## Independent CR

Blocking: none.

Should-fix: none after fixes.

Nits: none.

Review fixes applied:
- removed caller-controlled capability overrides from verification/install and isolated the internal capability registry from the frozen public snapshot
- cross-validated top-level requirements against Adapter identity/capabilities, `runtime-capability` operation bindings, Rule module identities/requirements, and quality-module bindings
- added regression coverage for forged capability support and capability stripping across all embedded sources

Verified:
- focused Harness tests: 6 pass / 24 assertions
- full deterministic suite: 468 pass
- typecheck, doctor, asset build, npm pack shape, Windows PE/Linux ELF builds, and diff check passed
- the original NN prerelease still reports exactly `prism-agent/delegation@1` and `quality-guard/anti-ai-flavor@1` as unsupported
- tampered prerelease variants that hide either capability are rejected fail-closed

3aKHP:
## Bot Review follow-up

Addressed in `d3855c0`:

- validate per-kind Adapter operation binding shape; reject unknown built-in tools in `tool-group` and `interaction-tool` bindings while preserving intentionally unavailable `optional-tool` declarations
- document and enforce that explicit released Harness Agent allowlists use portable Vesicle built-in host tools, not project-local MCP or parent-only names
- remove the redundant install option field
- resolve host assets before the final file/symlink check
- normalize concurrent destination conflicts to the stable `already installed` error
- add an Adapter unknown-tool regression test

Not changed in this foundation PR:

- the two-line `errorMessage` helper remains local to each parser/verifier module; extracting a shared module would add indirection without changing behavior
- the SemVer pattern remains aligned with the NN `harness-manifest.schema.json` and `shared/prism-driver/scripts/lib.ts` validator; Vesicle should not unilaterally reject versions accepted by the producer contract. Strict SemVer grammar should be coordinated upstream first

Post-fix verification: focused Harness 7 pass / 25 assertions, full suite 469 pass, typecheck, doctor, npm pack shape, PE/ELF builds, real prerelease verification, independent CR, and `git diff --check` all pass.

PR #25, opened 2026-07-14
feat(assets): bundle verified V10 baseline

3aKHP:
## Summary

- replace the bundled legacy V9 asset tree with the verified Neural Narratology V10 Harness baseline and root manifest
- split Vesicle-owned prompts and generic Agent profiles into a separate host asset layer while preserving project/user overrides and managed-pack activation
- pin bundled Harness identity into new sessions, fail closed on incompatible legacy resumes, and align delegation, quality guard, CLI, packaging, docs, and tests
- preserve manifest byte hashes across Windows checkouts and harden malformed contract-delegation recovery

## Test Plan

- [x] `bun run typecheck`
- [x] `bun test` (533 pass, 0 fail)
- [x] `bun run doctor`
- [x] `bun run build:assets`
- [x] `bun run pack:check`
- [x] `bun run build:exe`
- [x] `git diff --check`
- [x] `core.autocrlf=true` isolated checkout (47 manifest hashes verified)
- [x] CI: Typecheck/tests, Linux ELF smoke, Windows PE smoke

## Notes / Follow-ups

- The opt-in real-provider E2E remained skipped because `BUN_E2E_REAL_PROVIDER` was not set; deterministic gate coverage passed in the full suite.
- Merge target is `develop`; independent CR is required before merge.


3aKHP:
Bot Review follow-up on current HEAD `6cc21df`:

- Fixed the malformed `spawn_agent` classification. Invalid JSON no longer consumes the single contract-delegation slot, and Agent tools now return an explicit valid-JSON error. Added an end-to-end regression covering a malformed call beside a valid `scene-writer` delegation.
- Replaced the hard-coded ETL profile probe in `bundledHarnessLayout` with manifest/assets/host-assets structure discovery. Strict manifest inventory verification remains responsible for precise missing/hash errors.
- Kept the `assets/manifest.json` lookup in the no-Harness diagnostic fallback. It is not part of the healthy bundled path; it lets `assets status` / `doctor` report a legacy or partial asset layout when no verified Harness runtime can be constructed.

The first Windows PE failure was separately fixed by enforcing LF for manifest-hashed text assets in `.gitattributes`; a `core.autocrlf=true` checkout verified all 47 hashes. Fresh CI on `6cc21df` is green across Typecheck/tests, Linux ELF smoke, and Windows PE smoke.

3aKHP:
Final independent CR on `a5f120d`: **0 Blocking, 0 Should-fix**.

Review fixes now cover:

- malformed and whitespace-normalized Agent profiles without consuming the contract delegation slot incorrectly
- Windows CRLF preservation for all manifest-hashed text assets
- fail-closed handling when `assets/`, `host-assets/`, or root `harness-manifest.json` is missing
- the same verified-Harness requirement across new turns, continuations, and TUI session resume

Verification on the final head:

- 533 tests pass; typecheck, doctor, pack check, executable build, and diff check pass locally
- direct Windows PE release-shape smoke resolves 47 bundled + 12 host assets
- incomplete-bundle smokes exit non-zero for missing assets, host assets, and root manifest
- fresh GitHub CI is green for Typecheck/tests, Linux ELF smoke, and Windows PE smoke

PR #27, opened 2026-07-14
fix(installer): remove global project lock

3aKHP:
## Summary

- stop Guided Setup from persisting a global first-project pointer
- launch projects from the invocation directory with `vesicle .`, explicit paths, or Explorer
- install the `vesicle` PATH alias and per-user Explorer directory actions

## Test Plan

- [x] `bun run typecheck`
- [x] focused Setup and installer tests
- [x] native Windows installer compile and install/uninstall smoke
- [x] `git diff --check`


3aKHP:
## Known issues / follow-up memo

Items 1-3 still require triage before deciding whether they block this PR. Item 4 is explicitly non-blocking for this PR.

1. **Installed command name mismatch, blocker status pending:** the command currently available in actual Windows testing is `prism-vesicle`, not the documented `vesicle`. Follow-up should either align the documentation with the shipped command or provide a reliable `vesicle` alias.
2. **Setup rendering outside fullscreen, blocker status pending:** some Guided Setup screens show overlapping or mixed characters when the terminal window is not fullscreen.
3. **Installer navigation and maintenance semantics, blocker status pending:** Guided Setup does not support returning to the previous step. When Prism Vesicle is already installed, launching the installer repeats the installation flow instead of presenting conventional **Reinstall / Repair / Uninstall** choices.
4. **Non-blocking for this PR; highest-priority subsequent fix:** the built-in `shell_exec` supports only PowerShell 7 on Windows, but PowerShell 7 is not installed by default on every Windows system. Follow-up should add an automatic fallback or a configurable shell interpreter.

This comment records current test findings. The blocking status of items 1-3 must be decided after triage rather than assumed here.

PR #28, opened 2026-07-15
feat(shell): add configurable interpreter profiles

3aKHP:
## Summary

- Follow up on [PR #27 memo item 4](https://github.com/3aKHP/prism-vesicle/pull/27#issuecomment-4970196478) by removing the hard PowerShell 7 dependency from Windows `shell_exec`.
- Add host-owned `auto`, `posix-sh`, `powershell-7`, `windows-powershell-5.1`, `cmd`, and `git-bash` interpreter profiles.
- Resolve the executable before approval and bind the profile id, absolute path, and runtime policy version into the approved plan. Explicit unavailable profiles fail closed without crossing command dialects.
- Normalize PowerShell and CMD output to UTF-8, adapt model-visible guidance to the resolved dialect, and surface the interpreter in Doctor, permission prompts, and process cards.
- Preserve existing shell settings during Guided Setup and update the Windows CI lane, configuration example, architecture guidance, and bilingual user documentation.

## Platform behavior

- Windows `auto` prefers PowerShell 7 and falls back only to Windows PowerShell 5.1.
- Linux/WSL `auto` and explicit `posix-sh` use `/bin/sh`; Windows-only profiles are unavailable.
- Automatic resolution does not inspect the user login shell or load Bash, Zsh, PowerShell, or Git Bash profile files.
- When a configured interpreter is unavailable, new `shell_exec` launches are removed while `shell_output` and `shell_stop` remain available for existing tasks.

## Verification

- `bun run typecheck`
- `bun test` (574 pass, 0 fail)
- `bun run doctor`
- `bun run pack:check` (263 files)
- `bun run build:assets`
- `bun run build:exe` (Linux ELF and Windows PE)
- Real Windows execution smoke for PowerShell 7, Windows PowerShell 5.1, CMD, and Git Bash with Chinese UTF-8 output
- Simulated missing PowerShell 7 fallback to Windows PowerShell 5.1
- `git diff --check`

The final Inno Setup installer compile is left to the Windows CI lane because `ISCC.exe` is not installed locally.


3aKHP:
## Independent CR follow-up

The independent review found no Blocking issues and one Should-fix: the git-bash profile accepted any bash.exe on PATH while advertising Git for Windows semantics.

Addressed in 928515d:
- require Git Bash candidates to match the Git for Windows bin or usr\bin layout and installation-root cmd\git.exe;
- reject unrelated PATH Bash implementations such as MSYS2;
- retain a positive PATH fallback for user-local Git for Windows installs.

The remaining automated-review notes were re-checked and do not require code changes:
- JavaScript default parameters are evaluated at call time, not module import time;
- legacy PowerShell plans fail closed at the exact-plan hash comparison and are never replayed;
- loading permission settings before the first YOLO turn is required to honor the configured shellInterpreter.

Verification on head 928515d:
- bun run typecheck
- bun test: 576 pass, 0 fail
- bun run doctor
- bun run pack:check: 263 files
- bun run build:assets
- git diff --check
- CI: Typecheck and tests, Linux ELF smoke, and Windows PE/interpreter/installer smoke all passed.

PR #33, opened 2026-07-15
fix(tui): prevent terminal text overlap

3aKHP:
## Summary

- measure shared TUI truncation and padding by terminal display columns, and disable implicit wrapping on fixed-height rows
- budget Guided Setup, Permission, Rewind, and YOLO panels so descriptions and decision controls remain on separate visible rows
- add CJK, compact-terminal, long-command, rewind-warning, and narrow YOLO regression coverage
- refresh AI collaborator routing for the reorganized ignored local documentation area

## Test Plan

- [x] `bun run typecheck`
- [x] `bun test` (585 tests)
- [x] `bun run doctor`
- [x] `bun run build:exe`
- [x] `git diff --check`

## Notes / Follow-ups

- The real-provider E2E remains skipped because this change is limited to local Setup and TUI rendering.


3aKHP:
Review follow-up applied in ed0e651:

- filled the YOLO panel, restored the wide hint, and added a compact narrow hint
- bounded repeated-space continuation, narrow Permission warning/detail/feedback rows, and Rewind restore errors
- aligned width measurement with Bun/OpenTUI and made truncation, wrapping, composer movement/deletion, and cursor rendering grapheme-safe
- kept Setup Review compact through heights 24-26 so every action remains visible
- documented why gate summaries retain character-boundary wrapping

Independent CR then rechecked the full updated diff and reported no remaining Blocking or Should-fix findings.

Local verification on the updated head:
- bun run typecheck
- bun test: 589 passed
- bun run doctor
- bun run build:exe
- git diff --check

PR #34, opened 2026-07-15
chore(ci): move actions to Node 24 runtimes

3aKHP:
## Summary

- move checkout, artifact, setup-node, and GitHub Release actions onto Node 24 runtime majors
- keep Bun 1.3.14 as the project CI runtime and retain explicit Node 24 for npm Trusted Publishing
- lock the action-version contract in workflow tests and document the CI/runtime split

## Test Plan

- [x] `bun test tests/release-workflow.test.ts`
- [x] `bun run typecheck`
- [x] `bun test` (590 pass)
- [x] `bun run doctor`
- [x] `git diff --check`

## Notes / Follow-ups

- `actions/download-artifact@v8` fails on artifact digest mismatch by default.
- `oven-sh/setup-bun@v2` remains unchanged because it already runs on Node 24.


3aKHP:
Independent CR completed.

- Blocking: none
- Should-fix: one finding addressed in `015440b` by making `digest-mismatch: error` explicit and enforcing it in the workflow contract test
- Re-review: Blocking and Should-fix are clear
- Verified upstream manifests: checkout v7, upload-artifact v7, download-artifact v8, setup-node v7, setup-bun v2, and action-gh-release v3 all declare Node 24 runtimes
- Latest CI: source/package, Linux release smoke, and Windows installer release smoke all passed

PR #35, opened 2026-07-16
fix(quality): bind rewrites to artifact post-images

3aKHP:
## Summary

- derive Output Quality Guard artifact targets only from successful create/write/replace/append `FileToolEvent` results
- persist per-path mutation ids, complete post-image hashes, byte counts, and rejected-hash history across permission pauses, cancellation, and restart
- reread every current UTF-8 post-image through the shared writable-path and symlink guards before delivery
- keep blocking paths independent so clean completion prose or another clean file cannot clear an unchanged bad artifact
- check same-response mutation plus gate rounds only after the mutation succeeds

## Why

The previous Runtime rewrite loop accumulated mutation arguments as strings and cleared them after requesting a rewrite. A model could leave a bad file unchanged, reply with a clean completion summary, and cause the Guard to record `rewrite -> pass`. Partial replace/append checks also inspected only the argument fragment instead of the complete resulting file.

This PR implements PR 1 from the local Output Quality Guard plan: target/post-image correctness. It intentionally keeps the existing detector rules, two-round rewrite budget, and exhausted delivery policy.

## Compatibility

- legacy `candidateParts` session and permission state remains readable
- new durable state stores target metadata rather than duplicate artifact contents
- unknown or malformed target metadata is rejected
- Rule Pack and Harness capability contracts are unchanged

## Not In Scope

- durable exhausted/inconclusive user outcomes and TUI decision UI
- interactive TUI quality-resume entry points
- new deterministic rules, document metrics, or Semantic Judge

## Verification

- `bun run typecheck`
- `bun test` (`597 pass`, `0 fail`)
- `bun run doctor` (`Missing: none`)
- `git diff --check`
- focused quality/file/session/permission suite (`80 pass`, `0 fail` before the final full run)


3aKHP:
Independent CR completed and re-reviewed at `c0f09cc`.

- Blocking: none
- Should-fix: none remaining
- Initial Should-fix findings addressed:
  - Scene Writer now evaluates successful artifact mutations from complete guarded post-images
  - Weaver-Orch target classification is consistently prose-only
  - artifact cancellation/restart and legacy `candidateParts` have end-to-end Guard coverage
- Local verification after fixes: `bun run typecheck`, `bun test` (`600 pass`, `0 fail`), `bun run doctor` (`Missing: none`), and `git diff --check`
- Independent focused re-review: `44 pass`, `0 fail`

Merge remains gated on the refreshed PR CI run.

PR #39, opened 2026-07-17
feat(quality): add semantic judge benchmark runner

3aKHP:
## Summary
- add a developer-only, explicit-live Semantic Judge benchmark runner
- persist resumable, privacy-safe JSONL measurements with budget reservations and Wilson/slice reports
- document the PR 6B evidence boundary and add cancellation, resume, cap, and three-protocol tests

## Verification
- bun run typecheck
- bun test
- bun run doctor
- bun run build:assets
- bun run pack:check
- bun run pack:smoke
- git diff --check

No real provider calls were made. The runner cannot enable semantic blocking.


3aKHP:
Addressed automated CR findings in 4274202:
- valid blinded results no longer enter failureCaseIds; covered by regression
- normal rate waits remove their abort listeners
- rule-set comparison sorts each input once

Focused verification: bun run typecheck; bun test tests/quality-benchmark.test.ts; git diff --check. Full PR CI has been re-triggered.

PR #41, opened 2026-07-18
fix(quality): freeze benchmark Judge timeout

3aKHP:
## Summary
- add a bounded judgeTimeoutMs field to frozen benchmark plans
- pass the deadline only to benchmark Judge calls and include it in reports and plan identity
- document the Runtime/benchmark timeout boundary and cover timeout propagation plus invalid plans

## Verification
- bun run typecheck
- bun test
- bun run doctor
- bun run pack:check
- git diff --check


3aKHP:
Addressed the low-severity review finding in e5f0fbb: the timeout propagation test now explicitly keeps repeatsPerCase at one. Focused test and typecheck pass.

PR #44, opened 2026-07-19
feat(stage): add consumer narrative engine

3aKHP:
## Summary
- add the released Stage Harness profile, prompt, context template, and verified manifest refresh
- start frozen Module A/B Stage sessions through `/stage`, with no model-visible tools or gates
- preserve Stage bootstrap metadata on resume and add `stage.prose` quality observation plus explicit Judge rewrite support

## Verification
- `bun run typecheck`
- `bun test` (672 passed)
- `bun run doctor`
- `bun run build:assets`
- `git diff --check`


3aKHP:
Independent CR completed and all findings were fixed in dba6f26.

- aligned `runtime-packet` validation with the released Stage Neural Chain and consumer HUD variant
- classified interaction calls as interactive only when they are in the effective model-visible tool surface, so Stage rejects them as unavailable
- made `/new` leave Stage for ETL because a fresh Stage narrative requires `/stage` bootstrap

Reverified: `bun test` (675 passed), `bun run typecheck`, `bun run doctor`, and `git diff --check`.

3aKHP:
Follow-up review findings are addressed in c91676b.\n\n- Stage now rejects a missing `beat_map` defensively before parsing beat fields.\n- Unbootstrapped Stage turns fail before provider, Harness, or asset setup.\n- Stage bypasses MCP registry construction and receives an empty registry.\n- A Scenario card whose visible body starts with a character line now has a precise opening-paragraph diagnostic.\n\nReverified locally: `bun test` (676 passed), `bun run typecheck`, `bun run doctor`, and `git diff --check`.

PR #46, opened 2026-07-20
feat(harness): refresh the 10.1 rc.1 baseline and Stage host

3aKHP:
## Summary

- adapt the bundled Harness baseline to 10.1.0-rc.1 with verified prompt-context assets and manifest updates
- bootstrap Stage from frozen raw cards and render prose-first packets in the TUI
- improve slash argument completion plus artifact and Stage file selection ergonomics
- synchronize public documentation, asset contracts, and focused regression coverage

## Validation

- `bun run typecheck`
- `bun test` (710 pass, 0 fail)
- `bun run doctor`
- `git diff --check origin/develop...HEAD`
- manual 120-column TUI smoke for artifact sidebar focus


src/tui/widgets/Message.tsx
```diff
@@ -37,7 +42,16 @@ type MessageLike = {
...
+  message: MessageLike & { id?: string };
+  reasoningMode: string;
+  width: number;
+  agent?: AgentCardState;
+  stageSource?: boolean;
+  onStageToggle?: () => void;
+  streaming?: boolean;
+}) {
+  const renderer = useRenderer();
   const m = props.message;
```
3aKHP: **Correctness — high (regression vs develop).** `const m = props.message` captures the message object once at mount. The streaming preview is now rendered via `<Message message={{ content: props.streamingAssistant, ... }} />` (MessageStream.tsx:160), so `content={m.content}` compiles to a getter over a plain captured property — no SolidJS signal is tracked and the streaming body never grows past the first token batch. On `develop` this read `props.streamingAssistant` directly inside JSX and was reactive.

Fix: read the prop reactively inside a tracked getter — e.g. a `createMemo(() => props.message)` referenced from JSX, or avoid the `const m` capture and reference `props.message.content` where the content is read. Affects every engine; Stage is doubly affected (`stageParsed` / `showStageProjection` at lines 70-76 also freeze, so the projection never activates mid-stream).

src/core/stage/bootstrap.ts
```diff
@@ -234,6 +182,12 @@ function sha256(value: string): string {
...
-function validationWarnings(...results: ValidationResult[]): string[] {
-  return results.flatMap((result) => result.warnings);
+function compatibilityWarnings(characterContent: string, scenarioContent: string, unclosedLogicMarker: boolean): string[] {
+  const warnings: string[] = [];
+  if (!/^\uFEFF?---\r?\n/.test(characterContent)) warnings.push("Module A has no YAML frontmatter; its supplied text was frozen unchanged.");
+  if (!/^\uFEFF?---\r?\n/.test(scenarioContent)) warnings.push("Module B has no YAML frontmatter; its supplied text was frozen unchanged.");
+  if (!scenarioContent.includes("<!--")) warnings.push("Module B has no optional logic content; Stage will continue from visible material.");
+  if (unclosedLogicMarker) warnings.push("Module B has an unclosed HTML comment marker; it was kept in visible content unchanged.");
+  if (characterContent.trim().length === 0 || scenarioContent.trim().length === 0) warnings.push("An empty Stage input was frozen unchanged; add narrative material before relying on continuity.");
+  return warnings.slice(0, 3);
```
3aKHP, line 192: **Correctness / UX.** `warnings.slice(0, 3)` keeps the first three warnings, but the most actionable one — "An empty Stage input was frozen unchanged" — is pushed last (line 191). When both inputs are empty (or both simply lack frontmatter), warnings 1-3 (no Module A frontmatter, no Module B frontmatter, no `<!--`) crowd it out, so the user never learns the session started with empty character context and an empty opening. No test currently exercises empty inputs.

Fix: promote the empty-input warning to the front, or drop/raise the slice cap so that diagnostic always surfaces.

src/tui/widgets/Message.tsx
```diff
@@ -53,15 +67,54 @@ export function Message(props: { message: MessageLike; reasoningMode: string; wi
...
-      <box flexDirection="column">
+      <box
+        id={m.id}
+        flexDirection="column"
+        focusable={showStageProjection}
+        onKeyDown={(key) => {
+          if (!showStageProjection || !isStageMessageToggleShortcut(key)) return;
+          key.preventDefault();
+          key.stopPropagation();
+          props.onStageToggle?.();
```
3aKHP: **Correctness — plausible.** The Stage source-toggle shortcut now has two dispatch paths: this box-level `onKeyDown` (calls `props.onStageToggle()` → `toggleStageMessage`) and the global `useKeyboard` → `handleStageMessageKey` (input-routing.ts:124 → MessageStream.tsx:115-117, calls `toggleStageMessage(focused)`). When a stage box has focus and `focusedStageMessageId` is set (e.g. after a click, which sets it at MessageStream.tsx:43), both can deliver the same keypress and net to a no-op double toggle. `key.stopPropagation()` here may not stop the separate global `useKeyboard` hook.

Worth confirming the focus/dispatch model; if both paths can fire, gate one of them.

src/tui/views/MessageStream.tsx
```diff
@@ -17,30 +21,170 @@ export function MessageStream(props: {
...
+
+  function stageMessageAt(y: number): string | undefined {
+    return eligibleStageMessageIds().find((id) => {
+      const renderable = scrollbox?.getRenderable(id);
+      return renderable !== undefined && y >= renderable.screenY && y < renderable.screenY + renderable.height;
+    });
+  }
+
+  function handleStageMessageKey(key: TuiKeyEvent): boolean {
+    const ids = eligibleStageMessageIds();
```
3aKHP, line 109: **Performance.** `eligibleStageMessageIds()` runs `parseStageMessageContent` over every stage assistant message, and it's called at the very top of `handleStageMessageKey` before any key filter. `input-routing.ts:124` invokes this handler on every keystroke in composer mode, so typing in the composer re-parses the whole stage transcript per key. The same cost repeats in `stableAnchorIds` / `stageMessageAt` on mouse and toggle paths.

Fix: memoize the parsed result per message (e.g. a `createMemo` over `props.messages`) and/or bail early when the key isn't stage-related before doing any parsing.

src/tui/widgets/StageMessageContent.tsx
```diff
@@ -0,0 +1,28 @@
...
+import { renderMarkdownPlainText } from "../markdown-display";
+import { Show } from "solid-js";
+
+export function StageMessageContent(props: {
+  content: string;
+  messageId: string;
+  source: boolean;
+  streaming?: boolean;
+}) {
+  const parsed = () => parseStageMessageContent(props.content, props.messageId, props.streaming === true);
```
3aKHP: **Performance.** `parsed()` is a plain accessor, not a `createMemo`, and is called 2-3× per render in the fallback branch (lines 17, 18, 24) — each call re-walks the full content character by character. On top of that, `Message.tsx:70` already computes `stageParsed` from the same `m.content` and then passes `m.content` in here to be parsed again.

Fix: wrap `parsed()` in `createMemo`, and thread the already-parsed result in as a prop from `Message` so the content is walked once per render, not three times.

src/tui/widgets/ArgumentMenu.tsx
```diff
@@ -14,19 +14,28 @@ export type ArgumentMenuProps = {
...
 
   return (
     <box flexDirection="column">
       <For each={win().visible}>
         {(item, getIndex) => {
           const index = () => win().start + getIndex();
           const isSelected = () => index() === safeSelected();
-          const label = padDisplayEnd(truncateLine(item.label, LABEL_COLUMN), LABEL_COLUMN);
-          const detailBudget = Math.max(0, props.width - LABEL_COLUMN - 2);
+          const labelBudget = argumentMenuLabelBudget(props.width, hasDetails());
```
3aKHP, line 34: **Performance.** `hasDetails()` runs `props.items.some(...)` over the entire candidate list but is called inside the `<For>` per visible row (lines 34 and 35), so each render is O(N×M). The result is identical for every row, so it only needs to be computed once per render. With 200 artifact candidates (from `artifactCommandCompletion`) and 8 visible rows this is ~3200 redundant detail-field checks per render.

Fix: hoist `hasDetails` outside the `<For>` (or into a `createMemo`).

src/tui/command-completion-controller.ts
```diff
@@ -48,58 +45,65 @@ export function createCommandCompletionController(options: CommandCompletionCont
...
-        ]
-      : handles.filter((item) => {
-          const agent = options.agentCards().find((candidate) => candidate.handle === item.id);
-          return agent?.status === "queued" || agent?.status === "running";
-        });
-    return matchOptionItems(agentDraft.query, candidates);
+    let current = true;
+    setLoadedSourceKey(null);
+    setLoadedItems([]);
+    void draft.items().then((items) => {
```
3aKHP: **Performance.** This effect re-runs whenever `sourceKey` changes. For `/stage`, `sourceKey` flips from `stage:character` to `stage:scenario` when the user types the space after the first card (argument-completion.ts:324), so `listStageCardPaths` — a full directory walk of every writable root, up to 200 files — runs a second time for the identical file list.

Fix: cache or share the listing across the two Stage stages, since the path set is the same for character and scenario cards.

src/tui/stage-message-content.ts
```diff
@@ -0,0 +1,237 @@
...
+
+/**
+ * Recognizes only the Stage three-part packet. The returned HUD offsets point
+ * into the unchanged source string so the renderer can compact it safely.
+ */
+export function parseStageMessageContent(content: string, messageId: string, streaming = false): StageMessageContent {
+  const segments = splitMessageCommentSegments(content, messageId);
+  const neuralChain = segments.find((segment) => segment.kind === "comment" && segment.raw.includes("[!Neural Chain]"));
+  if (!neuralChain) {
+    const pendingCommentStart = streaming ? unfinishedCommentStart(content) : undefined;
```
3aKHP: **Performance + duplication.** When `streaming=true` and no `[!Neural Chain]` comment is present yet, the content is walked twice — once by `splitMessageCommentSegments`, then again by `unfinishedCommentStart`, which re-implements the same fence / inline-tick / HTML-comment state machine (lines 187-220 duplicate the walker at lines 27-79).

Fix: have `splitMessageCommentSegments` also return the unfinished-comment position so the second walk is eliminated. This also removes the hazard of the two scanners drifting apart on a fence/backtick rule change.

src/tui/commands/argument-completion.ts
```diff
@@ -111,6 +120,267 @@ export function completeAgentArgument(draft: AgentArgumentDraft, item: OptionIte
...
+  return match[2] ?? null;
+}
+
+function commandName(value: string): string | null {
+  const match = value.match(/^\/([^\s]+)/);
+  return match?.[1]?.toLowerCase() ?? null;
+}
+
+function splitTokens(value: string): { values: string[]; trailingSpace: boolean } {
+  const values = value.match(/(?:[^\s"]+|"[^"]*")+/g)?.map((part) => part.replace(/^"|"$/g, "")) ?? [];
```
3aKHP, line 375: **Cleanup — reuse.** The new `splitTokens` helper duplicates the exact tokenizing regex + replace still inlined in the `/stage` run handler (builtin.ts:66): the same `match(/(?:[^\s"]+|"[^"]*")+/g)` and `replace(/^"|"$/g, "")`. The helper was added but the existing call site wasn't refactored to use it.

Fix: call `splitTokens` from the stage run handler so any quoting/escaping fix lives in one place — otherwise `/stage` completion and `/stage` execution can parse arguments differently.

src/tui/views/MessageStream.tsx
```diff
@@ -17,30 +21,170 @@ export function MessageStream(props: {
...
+    queueMicrotask(() => restore(false));
+  }
+
+  function stableAnchorIds(): string[] {
+    return props.messages.flatMap((message, index) => {
+      const id = messageId(message, index);
+      if (message.role !== "assistant" || message.engine !== "stage") return [id];
+      const parsed = parseStageMessageContent(message.content, id);
+      if (message.kind !== "stage-bootstrap-opening" && !parsed.hud) return [id];
+      return parsed.segments.flatMap((segment) => segment.kind === "comment" ? [segment.id] : [segment.id]);
```
3aKHP: **Cleanup — dead code / likely unfinished.** The ternary `segment.kind === "comment" ? [segment.id] : [segment.id]` has identical arms, so the condition is dead — it's equivalent to `.map((segment) => segment.id)`. Given `stableAnchorIds` exists to preserve scroll position across the source/projection toggle, and the sibling `normalStageRenderParts` deliberately distinguishes concealed-comment anchors from markdown, this looks like an unfinished anchor differentiation rather than intentional behavior.

Fix: either restore the intended comment-vs-markdown anchor distinction, or collapse to `.map((s) => s.id)` so the intent is unambiguous.

3aKHP (commented):
Code review over the `develop..HEAD` diff — 10 findings, detail in the inline comments. Grouped by severity below. The headline item is #1: a SolidJS reactivity regression that freezes the live streaming preview at the first token batch.

## Correctness
1. **`src/tui/widgets/Message.tsx:55`** — `const m = props.message` loses reactivity; the new streaming `<Message>` wrap (MessageStream.tsx:160) captures the content once, so the streaming body never grows. Regression vs `develop`, which read `props.streamingAssistant` directly. **High, user-visible in normal use.**
2. **`src/core/stage/bootstrap.ts:192`** — `warnings.slice(0, 3)` drops the actionable "empty Stage input was frozen unchanged" warning whenever three cosmetic warnings also fire.
3. **`src/tui/widgets/Message.tsx:93`** — Stage source-toggle shortcut is wired through two dispatch paths (box `onKeyDown` + global `useKeyboard`); plausible double-toggle that nets to a no-op.
4. **`src/tui/views/MessageStream.tsx:79`** — dead ternary in `stableAnchorIds`; likely an unfinished comment-vs-markdown anchor differentiation.

## Performance
5. **`src/tui/views/MessageStream.tsx:100`** — `handleStageMessageKey` parses every stage message on every composer keystroke (no memo, no early-bail).
6. **`src/tui/widgets/StageMessageContent.tsx:12`** — `parsed()` not memoized → 2-3× content walk per render, plus a second parse in `Message.tsx`.
7. **`src/tui/widgets/ArgumentMenu.tsx:34`** — `hasDetails()` called per row inside `<For>` → O(N×M) per render.
8. **`src/tui/command-completion-controller.ts:74`** — `listStageCardPaths` runs twice per `/stage` (character→scenario stage flip) for the same file list.

## Cleanup
9. **`src/tui/commands/argument-completion.ts:376`** — new `splitTokens` duplicates the regex still inlined in `builtin.ts:66`; refactor the call site.
10. **`src/tui/stage-message-content.ts:89`** — double content walk in the streaming parse path; two copies of the fence/tick/comment scanner to maintain.

---

Not filed: the "hardcoded Stage HUD markers in TS" item follows an established pattern (the same marker list already lives in `src/core/validators/index.ts:344`) rather than introducing a new CLAUDE.md violation, so I left it out.

Happy to open fixes — #1 is a localized reactivity fix and the only finding that's user-visible in normal use.

src/tui/widgets/Message.tsx
3aKHP: Fixed in a2abea2. Message now reads  directly in tracked expressions, so the streaming prop is no longer captured as a mount-time object; Stage parsing/projection is memoized from those reactive reads.

src/tui/stage-message-content.ts
```diff
@@ -0,0 +1,237 @@
...
+  const lineEnd = content.indexOf("\n", start);
+  if (lineEnd < 0) return { text: content.slice(start).replace(/\r$/, ""), end: content.length };
+  return { text: content.slice(start, lineEnd).replace(/\r$/, ""), end: lineEnd + 1 };
+}
+
+function sliceSegment(segment: MessageCommentSegment, start: number, end: number, ordinal: number): MessageCommentSegment {
+  return { ...segment, id: `${segment.id}:visible:${ordinal}`, raw: segment.raw.slice(start - segment.start, end - segment.start), start, end };
+}
+
+function skipWhitespace(content: string, index: number): number {
```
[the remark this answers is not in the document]
3aKHP: Fixed in a2abea2. A recognized Neural Chain is retained as parser metadata even when the HUD is malformed, so Stage projection still conceals complete HTML comments while leaving malformed HUD text visible for graceful degradation.

src/core/stage/bootstrap.ts
3aKHP: Fixed in a2abea2. The empty-input warning is now inserted first, preserving it inside the three-warning cap, with a regression test for two empty card files.

src/tui/views/MessageStream.tsx
3aKHP: Fixed in a2abea2. Non-Stage keys now return before consulting Stage metadata, and Stage messages are parsed once in a  over the transcript instead of once per key/path.

src/tui/command-completion-controller.ts
3aKHP: Fixed in a2abea2. Dynamic completion loads are cached by logical source; both  and  share one  promise/listing while preserving per-stage selection keys.

src/tui/stage-message-content.ts
3aKHP: Fixed in a2abea2. Stage comment parsing is now a single pass that returns complete segments plus any pending comment start, eliminating the second streaming walk and scanner drift risk.

src/core/stage/bootstrap.ts
```diff
@@ -234,6 +182,12 @@ function sha256(value: string): string {
   return createHash("sha256").update(value).digest("hex");
 }
 
-function validationWarnings(...results: ValidationResult[]): string[] {
-  return results.flatMap((result) => result.warnings);
+function compatibilityWarnings(characterContent: string, scenarioContent: string, unclosedLogicMarker: boolean): string[] {
+  const warnings: string[] = [];
```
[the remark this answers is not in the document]
3aKHP: Fixed in a2abea2. The empty-input diagnostic is promoted ahead of the bounded compatibility warnings; an empty-card regression test now protects this case.

src/tui/widgets/ArgumentMenu.tsx
3aKHP: Fixed in a2abea2.  is now a Solid , so all visible rows reuse one candidate-list scan.

src/tui/widgets/Message.tsx
3aKHP: Fixed in a2abea2. Stage keyboard dispatch is now owned only by the global MessageStream handler; the box-level  path was removed. Mouse handling prevents/stops the handled click before the scrollbox path.

src/core/stage/bootstrap.ts
```diff
@@ -123,101 +144,28 @@ export async function stageSourceDrift(rootDir: string, bootstrap: StageBootstra
...
-    "module_b.first_line": firstLine(scenario.body),
-    "module_b.scene_premise": commentSection(scenario.body, "Scene Premise"),
-    "module_b.surface_emotion": commentField(scenario.body, "Surface emotion"),
-    "module_b.tension_source": commentField(scenario.body, "Tension source"),
-    "module_b.active_lens": commentField(scenario.body, "Active lens"),
-    "module_b.identity": commentField(scenario.body, "Identity"),
-    "module_b.immediate_goal": commentField(scenario.body, "Immediate goal"),
-    "module_b.first_beat.label": beatField(scenario.frontmatter, "label"),
-    "module_b.first_beat.tension_target": beatField(scenario.frontmatter, "tension_target"),
-    "module_b.first_beat.variant_config": beatField(scenario.frontmatter, "variant_config"),
```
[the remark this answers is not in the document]
3aKHP: Fixed in a2abea2. Stage template fence matching now accepts CRLF with  on both fence boundaries.

src/tui/stage-message-content.ts
```diff
@@ -0,0 +1,237 @@
...
+  let index = 0;
+  let inlineTicks = 0;
+  let fence: { marker: "`" | "~"; length: number } | undefined;
+  while (index < content.length) {
+    if (isLineStart(content, index)) {
+      const lineFence = readFence(content, index);
+      if (lineFence && (!fence || (lineFence.marker === fence.marker && lineFence.length >= fence.length))) {
+        fence = fence ? undefined : lineFence;
+        index += lineFence.width;
+        continue;
```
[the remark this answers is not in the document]
3aKHP: Fixed in a2abea2. One fence/tick-aware scanner now returns both complete segments and the unfinished-comment offset; the duplicated  walker was removed.

src/tui/widgets/StageMessageContent.tsx
3aKHP: Fixed in a2abea2. MessageStream memoizes each Stage parse and threads the parsed value through Message into StageMessageContent, which no longer reparses content in each render branch.

src/tui/commands/argument-completion.ts
3aKHP: Fixed in a2abea2. Exported the shared  helper and reused it in  execution, so completion and dispatch now share the same quoting/token rules.

src/tui/views/MessageStream.tsx
```diff
@@ -17,30 +21,170 @@ export function MessageStream(props: {
...
               width={props.contentWidth}
               agent={message.agentRunId ? props.agents.find((agent) => agent.runId === message.agentRunId) : undefined}
-            />
-          ))}
+              expanded={() => message.stageSource === true}
+              onToggle={() => toggleStageMessage(id)}
+            />;
+          }}</For>
           <Show when={props.streamingReasoning.trim().length > 0 && props.reasoningMode !== "hidden"} fallback={<box height={0} />}>
             <ReasoningBlock content={props.streamingReasoning} streaming={true} mode={props.reasoningMode} width={props.contentWidth} />
```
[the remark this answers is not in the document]
3aKHP: Fixed in a2abea2. Removed the unused  prop and its full-transcript allocation at the App call site.

src/cli/commands/prompt-dump.ts
```diff
@@ -38,17 +39,16 @@ export async function runPromptDump(args: string[]): Promise<void> {
...
-    await printShape(profile, bundle, systemPrompt, permissions.shellExec, permissions.shellInterpreter);
+    await printShape(profile, bundle, systemPrompt, permissions.shellExec, permissions.shellInterpreter, harness?.assets);
     return;
   }
 
-  await printFullDump(profile, bundle, systemPrompt, permissions.shellExec, permissions.shellInterpreter);
+  await printFullDump(profile, bundle, systemPrompt, permissions.shellExec, permissions.shellInterpreter, harness?.assets);
 }
 
 type ParsedArgs = { ok: true; value: { engine: EngineId; shapeOnly: boolean } } | { ok: false; error: string };
```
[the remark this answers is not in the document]
3aKHP: Fixed in a2abea2. Removed the obsolete host-contract name set, scan loop, return field, and empty prompt-dump output line; prompt contract tests now assert only the effective model-visible surface.

src/tui/views/MessageStream.tsx
3aKHP: Fixed in a2abea2. Collapsed the identical ternary to a direct segment-id map; anchor intent is now explicit and behavior is unchanged.

Issue #54, opened 2026-07-21
`assets` command leaks a raw stack trace on a corrupt Harness manifest (should fail closed cleanly, like launch/quality)


3aKHP:
## Problem

When `harness-manifest.json` is missing or corrupt, `vesicle assets …` throws an **uncaught** error and Bun dumps a ~130 KB stack trace to stderr before exiting 1. The fail-closed contract itself is fine (it exits non-zero and does not proceed with corrupt data); the problem is the user-facing surface — a wall of stack instead of a one-line error.

The CLI already has the right pattern for this in two sibling commands:

- `launch` → `launchProjectArgumentOrReport` wraps the call in `try/catch`, prints `error.message`, sets `process.exitCode = 1`. (`src/cli/main.ts`)
- `quality benchmark` wraps `runQualityBenchmarkCommand` in `try/catch`, `console.error(message)`, `process.exitCode = 1`. (`src/cli/main.ts`, `case "quality"`)

But `case "assets"` calls `await runAssetsCommand(...)` with no guard, so a throw propagates as an uncaught exception.

## Reproduction

```bash
bun run build:exe linux
mkdir -p smoke/release smoke/project
cp prism-vesicle smoke/release/prism-vesicle
cp -r assets smoke/release/assets
cp -r host-assets smoke/release/host-assets
printf '{ not a valid manifest }' > smoke/release/harness-manifest.json
( cd smoke/project && ../release/prism-vesicle assets status )
# exit 1, ~130 KB stack trace on stderr beginning with:
#   throw new Error(`Cannot read Harness manifest at ${path}: ${errorMessage(error)}`);
```

`doctor` and `debug markdown-runtime` are unaffected (they either tolerate a missing manifest or don't read it).

## Impact

- **Severity: low.** Fail-closed holds; no corrupt data is acted on.
- **UX: poor.** A user who corrupts/loses the manifest sees a stack dump instead of `Cannot read Harness manifest at <path>: <reason>`.
- **Consistency:** `assets` is the only user-facing command path left that leaks an uncaught stack.

## Suggested approach

Wrap the `assets` dispatch in `src/cli/main.ts` the same way `quality` already does:

```ts
case "assets": {
  const { runAssetsCommand } = await import("./assets");
  try {
    await runAssetsCommand(invocation.args.slice(1));
  } catch (error) {
    console.error(error instanceof Error ? error.message : String(error));
    process.exitCode = 1;
  }
  break;
}
```

[...]

PR #55, opened 2026-07-21
test: restructure test layer (directories, strict acceptance, CLI/binary journeys)

3aKHP:
## Summary

Phased audit and refactor of the test layer (plan: `dev/docs/working/TEST_LAYER_AUDIT_AND_REFACTOR_PLAN.md`). **Tests-only** — no production behavior change — plus the contributor-facing scripts and docs that track them.

- **Phase 1 (`b0c2f0b`)** — false-green hygiene: `e2e-gate` moved out of default discovery (`.acceptance.ts`), platform early-returns → `test.skipIf`, removed `build-assets` restatement, narrowed release/installer/cli static guards.
- **Phase 2–4 (`b31cfdd`)** — shared support (`tests/support/async/eventually.ts`, `tests/support/providers/sse.ts`) and split of the five aggregation files (`quality-runtime`, `agent-loop`, `session`, `harness-delegation`, `tui`) into `tests/integration/<domain>/` with co-located fixtures. Test bodies moved verbatim; assertions unchanged.
- **Phase 5 (`c19d23f`)** — migrated all 84 flat `tests/` root files into `tests/<level>/<domain>/` (unit / component / integration / contract / acceptance); added layered scripts (`test:unit/component/integration/contract/deterministic/acceptance:provider`); `e2e-gate` → `tests/acceptance/provider/`.
- **Phase 6 (`ce3f6be`)** — split acceptance into a connectivity smoke + strict ETL gate; both use `test.skipIf(!precondition.ok)` so a missing env/credentials reports as **skip, not pass**. Structured secret-free summaries.
- **Phase 7 A+D (`47edf2d`)** — first automated paths from a real user entry: CLI source journey (`tests/integration/cli/commands.test.ts` — command surface, exit codes, config isolation, error contracts) and setup-to-launch journey (`tests/integration/cli/setup-to-launch.test.ts` — Setup writes config → real `doctor` loads it).
- **Phase 7 B (`ee190ab`)** — standalone binary fail-closed smoke (`scripts/smoke-binary.ts`): embedded tree-sitter worker, asset resolution beside the executable, doctor parity, and corrupt-manifest fail-closed; wired into both Linux and Windows release lanes.

**Result**: `bun test` 717 → **726 pass / 5 skip / 0 fail / 129 files** (was 88 flat files / 710 pass). Acceptance lane without env: **0 pass / 2 skip** (was a false-green 1 pass). Binary smoke: 4/4.

Phase 7 C (native TUI PTY) and E (provider protocol matrix) are intentionally deferred — see Notes.

## Test Plan

- [x] `bun run lint`
- [x] `bun run typecheck`
- [x] `bun test` (726 pass / 5 skip / 0 fail / 129 files)
- [x] `bun run doctor` (Missing: none)
- [x] `bun run test:acceptance:provider` (0 pass / 2 skip without env — no false green)
- [x] `bun run smoke:binary` (4/4; requires `bun run build:exe linux`)

Not exercised: real-provider acceptance (needs credentials), Windows binary smoke (WSL builds Linux only — cross-platform env hardened; first Windows CI run will confirm).

## Notes / Follow-ups

- **Phase 7 C (native TUI PTY smoke) deferred**: a PTY driver is a new platform-specific dependency with real flakiness risk on Windows; value is genuine but cost/risk exceeds alpha-stage benefit. Continuing manual dogfood; gap recorded in the plan doc.
- **Phase 7 E (provider protocol matrix) deferred**: Phase 6 laid the smoke/strict-gate foundation; remaining work is breadth across OpenAI-compatible / Anthropic / Gemini, which needs credentials + spend + tolerates non-determinism. Event-driven: add a protocol's matrix when its adapter changes.
- **CHANGELOG not updated**: this is test-layer-internal (no runtime / behavior / tool-surface / config change); the `test` script delegates to `test:deterministic` but plain `bun test` stays equivalent. Flag if you want an entry anyway.
- **Finding (filed separately as #54)**: the `assets` command leaks a raw ~130 KB stack trace on a corrupt Harness manifest (`src/cli/main.ts` `case "assets"` has no try/catch, unlike `launch`/`quality`). Fail-closed (exit 1) holds; the UX leak is a separate production fix, kept out of this test-only PR.
- **Independent CR**: pending (the implementer cannot self-review); run before merge per `docs/dev/WORKFLOW.md`.


3aKHP:
Thanks @KHPilot[bot] for the review. Addressed per-finding (`0466a27`):

**F1 (`./` prefix on `test:acceptance:provider`) — not applied.** The `./` is load-bearing, not cosmetic. Bun treats a bare non-`.test.`/`.spec.` file argument as a *filter*, not a path, so dropping it makes the lane search 1228 files and match 0. Bun's own diagnostic requests the `./` form:

```
$ bun test tests/acceptance/provider/e2e-gate.acceptance.ts
1228 files were searched
note: To treat the "tests/acceptance/provider/e2e-gate.acceptance.ts" filter as a path,
      run "bun test ./tests/acceptance/provider/e2e-gate.acceptance.ts"
```

The other scripts use bare paths because they target *directories* (always treated as paths) that contain `.test.*` files. Acceptance uses the `.acceptance.ts` suffix deliberately (excluded from default discovery — Phase 1), so it requires the explicit `./` path form. Leaving as-is.

**F2 (removed windows-installer `.ps1` guards) — documented.** Added a scope comment in `tests/contract/release/windows-installer.test.ts`: the `.ps1` runtime behavior is covered by the release-lane installer smoke, and the removed text-matching was a redundant test-of-the-test (Phase 1 decision). Documented rather than restoring assertions — the `.ps1`'s internal structure isn't a stable architecture contract worth re-asserting; structural breakage is caught at release time by the installer smoke that actually runs it.

**F3 (`REPO_ROOT = process.cwd()`) — fixed.** `scripts/smoke-binary.ts` now resolves `REPO_ROOT` from `import.meta.dir`, so it's invocation-directory-agnostic. Verified 4/4 from both the repo root and `/tmp`.

Verification this commit: `bun run lint`, `bun run typecheck`, `bun run smoke:binary` (4/4 from two cwds), `bun test tests/contract/release/windows-installer.test.ts` (4/4).

PR #63, opened 2026-07-23
feat(instructions): persistent instructions auto-loaded into the system prompt (Increment A)

3aKHP:
## Summary

Adds a host-owned **Persistent Instructions** layer: user-authored Markdown that customizes an Engine's workflow and survives new sessions, so a tester no longer has to ask the model to write a sub-workflow spec to a file and remind it to read it next session (and no longer has to copy files between working folders).

This is **Increment A** (auto-load + inspection). It implements the plan in `dev/docs/working/PERSISTENT_INSTRUCTIONS_IMPLEMENTATION_PLAN.md` with two scoping cuts agreed during review:

- **No `SystemPromptBlock` cross-surface refactor.** The plan proposed threading an ordered block type through ~11 continuation surfaces. The engine prompt stays a composed `string`; a single composer (`composeSystemPromptWithInstructions`) appends instruction blocks after the byte-identical engine prompt and is called only at the system-prompt construction sites. In-process continuations reuse frozen instruction blocks; a restart resume re-resolves current files once and freezes that result for the rest of the turn.
- **Naming aligned to `VESICLE.md` / `VESICLE.<engine>.md` at both scopes** (project root + user config beside `providers.yaml`), instead of `instructions.md` vs `VESICLE.md`.

Resolution is **replacement within a scope, composition across scopes**: an engine-specific file replaces the scope's general file (an empty file is an explicit empty override that suppresses fallback); the user file is followed by the project file; project wins on a direct conflict. Neither can override the Engine contract.

### What's wired
- New `src/core/instructions/` module: enum `{ scope, engine }` targets, fixed path derivation, fail-soft loader (fatal UTF-8 + BOM strip, regular-file + symlink guards), resolution algebra, 32 KiB combined budget (project retained when the pair overflows), envelope renderer, fingerprint + audit-record projection.
- Single composer injected into **all five** system-prompt construction sites: turn bootstrap, continuation context, Stage bootstrap, `/compact`, and the `/btw` snapshot resolver. Order is engine → instructions → Stage context.
- Live re-resolution on every top-level turn, resume, and confirmed engine switch (live config, not session identity). Session-start system record stores the resolution (targets/hashes/bytes/diagnostics; no full content).
- Inspection: `vesicle prompt shape` / `prompt dump --engine <id>` and a new `/instructions` command show the effective selection, provenance, byte budget, and diagnostics.

### Deferred (Increment B / follow-up)
Model-visible `read_instructions` / `update_instructions` tools (Tool Permission Runtime integration, optimistic concurrency, atomic write, previous-state backup), per-turn change-detection audit records, and `vesicle doctor` instruction reporting. Instruction files are authored with a text editor today. The capability boundary (no tool/permission/gate/validator/path-authority widening) holds regardless because the tool runtime enforces capabilities independently.

## Test Plan

- [x] `bun run lint`
- [x] `bun run typecheck`
- [x] `bun test` (844 pass / 5 skip / 0 fail)
- [x] `bun run doctor`
- [x] `bun run src/cli/main.ts prompt shape --engine etl`
- [x] `bun run src/cli/main.ts prompt shape --engine stage`
- [x] `git diff --check`
- [x] `bun run pack:check` (330 files)

New coverage: instruction resolution algebra (absent / general / engine-override / all-four / empty-override / fallback-suppression), validation (invalid UTF-8, directory target, project + user symlink skip, individual + combined budget, exact boundary), BOM-stripped hashing, fingerprint identity, audit-record shape; provider ordering fixture proving instructions reach the provider system authority in `engine → user → project` order and that an engine override replaces general; `/instructions` command rendering; session-start metadata capture.

## Notes / Follow-ups

- Independent CR completed; continuation freezing, terminal cleanup, and diagnostic-state findings were fixed with focused regressions.
- Real TUI + real-provider dogfood remains opt-in and was not used as a merge gate; provider-wire integration coverage verifies the composed instruction authority.


3aKHP:
Follow-up to the CR re-review (commit `30a84c9`).

**#2 (Blocking) — frozen within a turn, implemented properly this time.** The previous commit weakened the approved contract to "live at every boundary"; this commit implements the frozen-within-turn contract from the plan:

- New `src/core/agent-loop/instruction-context.ts`: an in-process `Map<sessionId, blocks>` holding the turn-start instruction blocks.
- `bootstrapTurn` freezes `composeInstructionBlocks(selection)` on every top-level turn; `loadContinuationContext` reuses the frozen blocks when present (engine prompt is still re-derived — deterministic — and the frozen blocks appended), so a tool call decided under one instruction set never continues under another after a mid-turn pause. When the cache is absent (process restart) it re-reads current disk — a restart is a resume boundary, not mid-turn.
- `runLoop` clears the snapshot on turn completion; a paused turn keeps it for its in-process continuation.
- STYLE.md / CHANGELOG / both user-manual pages now describe the frozen contract accurately (resolve at top-level turn / restart-resume / engine switch; frozen within a turn; edits take effect next turn).

Tested by a pause→edit→resume integration test: the resumed continuation sends the **turn-start** instruction text to the provider, not the mid-pause edit; a second test clears the snapshot (simulating restart) and asserts the continuation then re-reads disk.

**Should-fix — dedup.** The `instruction_warning` TUI notice is now deduped by `session + diagnostic fingerprint` (closure-scoped), so a persistent invalid file does not re-notify every turn; a changed diagnostic set re-notifies. Added a controller-level test asserting dedup against real TUI messages.

**Should-fix — symlink skip.** Replaced `skipIf(win32)` with a module-load symlink-capability probe, so Windows-with-developer-mode runs the real assertion and only a genuine `EPERM`/`ENOSYS` skips.

Verification: `bun run lint` (clean, no warnings), `typecheck`, `bun test` (839 pass / 5 skip / 0 fail), `doctor`, `prompt shape` etl+stage, `git diff --check`.

Still outstanding before merge: real TUI + real-provider smoke (write a `VESICLE.runtime.md`, run a turn that pauses, confirm the model sees the rule and that editing it mid-pause does not change the in-flight turn).

PR #64, opened 2026-07-23
feat(init): /init scans the project and seeds a project-scope VESICLE.md

3aKHP:
## Summary

`/init [notes]` — the companion to Persistent Instructions (#63). Instead of hand-writing the first `VESICLE.md`, it scans the project's writable roots and makes one no-tools provider call with a dedicated host prompt to draft a project-scope `VESICLE.md`, which takes effect on the next top-level turn and is meant to be reviewed/edited.

Per the feasibility review, this is **Option A** (compact-style standalone provider call), chosen over prompt-injection or a SubAgent because Vesicle's engines are domain-specific (an init task doesn't fit a runtime/etl workflow) and because the init prompt lives in the **Vesicle host-assets layer** (`host-assets/prompts/shared/init-project.md`, added to `bundledHostAssetPaths`) — so it needs **no new Harness from Neural-Narratology**. The pre-existing feasibility doc's `workspace/PROJECT.md` framing is superseded: `VESICLE.md` is the CLAUDE.md-analog and auto-injects for free.

### What's wired
- `src/core/init/scanner.ts`: bounded host-side scan of the writable roots (file tree + short text-file heads; notes an existing VESICLE.md).
- `src/core/init/service.ts`: `generateProjectInstructions` — resolve harness → read the init host prompt → scan → one provider call (system = init prompt, user = digest + optional notes, no tools) → write `<root>/VESICLE.md`. An existing file is backed up to `.vesicle/init-backups/VESICLE.md.previous` before being replaced (copy, so a failed write leaves the original).
- `/init` command (builtin.ts) mirrored after `/compact`: `afterAgentLoop`, cancellable, with `initProject` plumbed through `CommandContext`, the turn-controller options, the session-actions controller, and all three TUI controller sites.
- Host asset `host-assets/prompts/shared/init-project.md` (Vesicle-native, RP/workflow-oriented: project overview, recommended engine, card/scenario conventions, workflow/naming conventions, non-obvious gotchas; capability-boundary aware).

### Deferred
SubAgent/interactive multi-phase init (ClaudeCode NEW-style `AskUserQuestion`), user-scope targeting, and streaming the draft as it generates. `/init` is one-shot, project-scope, for MVP.

## Test Plan

- [x] \`bun run lint\`
- [x] \`bun run typecheck\`
- [x] \`bun test\` (850 pass / 5 skip / 0 fail)
- [x] \`bun run doctor\`
- [x] \`git diff --check\`
- [x] \`bun run pack:check\` (334 files)

New coverage: scanner (root listing with text heads, existing-VESICLE.md note, non-text head skip) and service with a stubbed provider (writes VESICLE.md from the response, backs up before overwriting, passes user notes into the request). The service test also proves the init host asset resolves via the bundled harness's host layer.

## Notes / Follow-ups

- Real TUI + real-provider smoke (run `/init` in an actual project, confirm the generated VESICLE.md is sensible and applies next turn) is the remaining pre-merge dogfood.
- An independent CR pass is expected (cross-module: new command, host-asset bundle, harness path).


3aKHP:
Follow-up to the independent workflow-CR (7-dimension multi-agent review) and Bot review (commit `170b1d9`). No Blocking issues were found — the implementation was verified correct in every reviewed dimension. All confirmed Should-fix/Nit items are addressed:

**Robustness (/init ↔ Persistent Instructions coupling surface):**
- Provider output is now sanitized: one outer ```` ```lang...``` ```` fence is stripped before writing `VESICLE.md` (mirrors compact's `formatCompactSummary`), so a mis-formatted response can no longer pollute every future session's system prompt.
- The target path is resolved through the PI resolver (`instructionFilePath`) instead of a hardcoded literal in both `service.ts` and `scanner.ts` — the two sides can no longer drift on the filename.
- The generated content is validated against the 32 KiB PI budget before writing; over-budget output is refused with a clear error instead of being silently dropped while the command claims success.
- Project-scope `VESICLE.<engine>.md` overrides that mask the general file are detected, and the success notice is qualified ("takes effect ... for engines without an override") so it never lies.
- A `/init` run that replaces an existing `.previous` backup now warns.
- `readHead` extension parse fixed (`lastIndexOf(".") === -1` returned the last path char via `slice(-1)`).

**Tests added** (the cross-feature round-trip is the key one — it locks the /init↔PI contract): /init→PI loader round-trip; fence-strip; empty/whitespace rejection (no file created, existing untouched, no backup); over-budget rejection; scanner per-root file cap + total byte cap; scanner path-confinement canary (providers.yaml/.env/.vesicle never reach the digest); `/init` added to the scheduler afterAgentLoop regression list.

**Nits/docs:** removed dead `| "initProject"` from `DecisionContinuationOptions`; added `/init` to the busy-behavior paragraph in both user-manual command references.

**Deferred (per review):** the `complete()` stream-loop signal check is a pre-existing shared gap (compact has it too); the abort signal is already threaded to the adapter. Backup directory name (`init-backups/` vs the PI plan's `instruction-backups/`) left as-is — no collision today, Increment B unbuilt.

Verification: lint (clean), typecheck, `bun test` (857 pass / 5 skip / 0 fail), `git diff --check`, `pack:check` (334 files).

PR #65, opened 2026-07-23
feat(instructions): read_instructions / update_instructions model tools (Increment B)

3aKHP:
## Summary

Increment B of Persistent Instructions (#63): the model can now read and manage `VESICLE.md` through two universal host tools on every **non-Stage** Engine. Per the plan (`dev/docs/working/PERSISTENT_INSTRUCTIONS_IMPLEMENTATION_PLAN.md` §9–11), with three scoping decisions:

- **Stage stays strictly tool-less.** VESICLE.md is a negative for the Stage consumer-RP engine, so the §9.3 "Stage gets universal controls" revision is **not** applied — both tools are added only in the non-Stage path of `resolveBuiltInTools` (Stage's early `return []` is untouched).
- **Backup dir `instruction-backups/`** (plan), `/init`'s `init-backups/` left as-is (no cross-PR churn).
- **Default permission request** in MANUAL/INERTIA this PR; custom unified-diff preview deferred (the security/correctness boundaries — CAS, atomic, backup, budget — all hold with the default request).

### What's wired
- `read_instructions` (observe): one exact `{ scope, engine }` target → content/size/hash/`selectedForActiveEngine`; absent → `exists=false`.
- `update_instructions` (mutate): write/delete one target via new `src/core/instructions/store.ts` — atomic write (temp + rename), `ifMatchSha256` CAS (`"absent"` or 64-hex; stale never overwrites), single previous-state backup, 32 KiB budget check across all affected Engines, empty-override support. Fixed `{ scope, engine }` enum target — no arbitrary path — writing outside the model-visible writable roots as a bounded host exception.
- Dispatch in `executeHostTool`; `activeEngine` + `sessionId` threaded into executor options. `permissions/policy.ts`: read=observe, update=mutate (existing Tool Permission Runtime — MANUAL/INERTIA pause, MOMENTUM/YOLO execute; no second approval layer).
- **Frozen-cache refresh**: a successful update refreshes the in-turn frozen snapshot (`instruction-context.ts`) so it applies on the next provider round of the same turn — the one mid-turn recompose, consistent with #63's frozen-within-turn.
- `InstructionToolEvent` metadata on `ToolResult`.
- **Layering**: `instruction-context.ts` (the frozen snapshot) moved `agent-loop` → `instructions/` (it's an instruction concept; avoids an `instructions → agent-loop` reverse dep the new tools file would create).

## Test Plan
- [x] `bun run lint`
- [x] `bun run typecheck`
- [x] `bun test` (868 pass / 5 skip / 0 fail)
- [x] `bun run doctor`
- [x] `git diff --check`
- [x] `bun run pack:check` (336 files)
- [x] `prompt shape --engine etl` shows both tools; `--engine stage` shows neither

New coverage (`tests/unit/core/instruction-tools.test.ts`): read present/absent/masked-by-override; write/overwrite/empty-override; delete + absent no-op; CAS (omitted/absent/stale/matching); combination-budget rejection (general-target write that oversizes with the other scope); affected-engines; symlink rejection; frozen-cache refresh; permission classification. Oracles are independent of implementation.

## Notes / Follow-ups
- A 7-dimension workflow CR will run next (this crosses tool, permission, path-authority, and session contracts).
- Custom MANUAL/INERTIA diff-preview is deferred (the default permission request surfaces the tool + arguments; CAS/atomic/backup/budget hold regardless).
- Real TUI + real-provider dogfood (ask the model to read/update VESICLE.md, confirm next-round effect + MANUAL approval) is the remaining pre-merge check.
- Note: an untracked `VESICLE.md` exists at the repo root — it is the maintainer's dogfood (谬因 project) instruction file, **not** part of this PR; left untracked deliberately.


3aKHP:
Follow-up to the 7-dimension workflow CR (3 Blocking + 8 Should-fix + 5 Nits; commit `1ba3702`). All confirmed findings fixed; verification green (874 pass / 5 skip / 0 fail, lint/typecheck/diff/pack clean).

**Blockings (all fixed):**
- **B1 — refresh was dead in the active loop.** The update executor wrote to the frozen-instruction Map, but `turn-loop` passed a fixed `args.systemPrompt` every round and never re-read it. Fix (Option 1 from the review discussion): `RunLoopArgs` now carries the base `enginePrompt`; after each tool round, non-Stage engines recompose the live `systemPrompt = enginePrompt + readFrozenInstructionBlocks(sessionId)`. Stage is skipped (no instruction tools; preserves character context). New integration test `instruction-update-refresh` drives a full update → next-round provider request and asserts the new content reaches the provider — the exact regression that hid B1.
- **B2 — activeEngine not threaded on the permission-resume path.** Added `activeEngine: context.profile.id` to the `executeHostTool` call in `permission-continuation` (the resume path), so MANUAL/INERTIA resume keeps the refresh + `selectedForActiveEngine` correct. `enginePrompt` threaded through all runLoop callers.
- **B3 — configuration.md self-contradiction.** Removed the stale "tools deferred" blockquote (en + zh-CN) that contradicted the new tools-available note.

**Should-fix (all fixed):** S1 affected-engines treats only absent overrides as fallback (matching resolveScope); S2 normalize trailing newlines (no read-modify-write drift); S3 success message conditionally claims "next provider round" only when the active engine is affected; S4 instruction tools added to the SubAgent `unsupportedChildToolNames` (§9.4); S5 removed "(deferred)" from STYLE.md. Tests S6 (delete-budget engine-specific), S7 (user-scope 0o600 + backup), S8 (engine-specific affectedEngines), N2 (backup priorState by content), N4 (structured affected-engines + override-exclusion).

**Not done (per review):** N1 parent-dir fsync (matches the config-writer convention; spec hedges to delete-only), N3/N5 (negative-path + MANUAL pause integration — structurally mitigated by generic permission-runtime tests).

Also: `.gitignore` the local dogfood `/VESICLE.md` at the repo root (per maintainer request).

Remaining pre-merge: real TUI + real-provider dogfood (ask the model to read/update VESICLE.md, confirm next-round effect + MANUAL approval + backup).

3aKHP:
Follow-up to the independent CR Agent (1 Blocking + 3 Should-fix; commit `1d5191e`). All fixed; 877 pass / 0 fail.

**Blocking — MANUAL/INERTIA resume used a stale system prompt for the first post-approval round.** `permission-continuation` built the continuation context (systemPrompt from the frozen cache) *before* the approved `update_instructions` ran, then passed that stale `context.systemPrompt` into `runLoop`; the turn-loop only recomposed after the next tool round. Fix: `runLoop` now calls `refreshLiveSystemPrompt` at its **start** (before the first round), so the resumed loop recomposes from the already-refreshed frozen snapshot. New regression test: `runPrompt(MANUAL) → needs_permission → resolvePermission(allow_once) → assert next request body carries the new instruction` — the exact case the workflow CR's MOMENTUM-only test missed.

**Should-fix:**
- `InstructionToolEvent` was silently dropped — `tool-result-recorder` now persists it in the session tool-record metadata + the `tool_result` event (and the event variant carries the field). before/after hash, target, affected engines are audited.
- First-create now records a `priorState:"absent"` backup (`.previous.json`, no payload) — every state change records a recoverable previous state.
- `read_instructions` no longer misreports invalid targets (bad UTF-8 / symlink / dir / read-error) as "does not exist" — returns `ok:false` with the diagnostic.

Tests added: MANUAL resume refresh, invalid-target read, absent prior-state backup.

PR #92, opened 2026-07-25
chore(deps): patch brace-expansion audit finding

3aKHP:
Bump the transitive `brace-expansion` override from **2.1.2 → 5.0.8** to clear **GHSA-mh99-v99m-4gvg**. The prior `2.1.2` pin (PR #48) patched GHSA-3jxr-9vmj-r5cp; the new advisory's range `<=5.0.7` covers all 2.x, so `5.0.8` is the only patched release.

### Known trade-off — bot MEDIUM, accepted openly

`5.0.8` is **API-incompatible** with the sole consumer. `minimatch@8.0.7` declares `"brace-expansion": "^2.0.1"` and calls `__importDefault(require("brace-expansion")).default(pattern)` — expecting v2's callable `module.exports = function`. `5.0.8` is `"type": "module"` and `require()` returns a namespace `{ expand, EXPANSION_MAX, … }` (not callable as a default). If this chain were ever executed, minimatch's brace handling would throw `TypeError: … is not a function`.

Why we keep it anyway:
- The chain (`@opentui/solid › babel-plugin-module-resolver › glob › minimatch › brace-expansion`) is **build-only and never executed by Vesicle** — Vesicle runs TS/TSX natively on Bun with no Babel phase; `babel-plugin-module-resolver` is `@opentui/solid`'s own build-time plugin, carried transitively but never loaded. Both the DoS advisory and this incompatibility sit in dead code.
- **No compatible patch exists:** every `brace-expansion` 2.x is vulnerable; `minimatch@10` uses `^4.0.0` → `4.0.1` (still `≤5.0.7`); nothing consumes `^5`. A dep bump cannot fix this without breaking the consumer.
- We keep the audit gate honest by pinning the patched version openly rather than suppressing the advisory (suppressing advisories via `bun audit --ignore` is off the table per repo convention).
- Alpha period — higher breakage tolerance.

**Revisit when** a consumer adopts `brace-expansion^5`, `@opentui/solid` drops the build dep, or Bun offers a finer resolution — then the override is removed. Tracked in a local `dev/docs/` note.

### Verification
- `bun audit`
- `bun run lint`
- `bun run typecheck`
- `bun test` (901 pass, 5 skip, 0 fail)
- `bun run doctor`
- `bun run pack:check`
- `git diff --check`

No tests added; the override touches only a transitive build dependency that Vesicle does not load.

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -22,7 +22,7 @@
   },
   "overrides": {
     "@babel/core": "7.29.7",
-    "brace-expansion": "2.1.2",
+    "brace-expansion": "5.0.8",
   },
   "packages": {
     "@babel/code-frame": ["@babel/code-frame@7.29.7", "https://registry.npmmirror.com/@babel/code-frame/-/code-frame-7.29.7.tgz", { "dependencies": { "@babel/helper-validator-identifier": "^7.29.7", "js-tokens": "^4.0.0", "picocolors": "^1.1.1" } }, "sha512-Aup7aUOfpbAUg2ROOJN6Iw5f9DMBlzu0mIkm/malLQFN/YQgO48wCj0Kxa3sEHJvPVFg7siR+qRInwXd2qhQKw=="],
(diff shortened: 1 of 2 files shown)


3aKHP:
@KHPilot[bot] — the MEDIUM is correct, thanks; verified it.

`require('brace-expansion@5.0.8')` returns the namespace `{ expand, EXPANSION_MAX, EXPANSION_MAX_LENGTH }` (not callable), while `minimatch@8.0.7` does `__importDefault(require(...)).default(pattern)`. So yes — a 3-major-version API gap; if this chain ran, minimatch's brace handling would throw `TypeError: ... is not a function`.

We're keeping the `5.0.8` override anyway, openly:
- The whole chain (`@opentui/solid › babel-plugin-module-resolver › glob › minimatch › brace-expansion`) is **build-only and never executed by Vesicle** — Bun runs TS/TSX natively, no Babel phase; the plugin is `@opentui/solid`'s own build-time dep, carried transitively but never `require`d. Both the DoS advisory and this incompatibility sit in dead code.
- **No compatible patch exists** — every `brace-expansion` 2.x is vulnerable; `minimatch@10` → `^4.0.0` (4.0.1, still ≤5.0.7); nothing consumes `^5`. So a bump can't fix this without breaking the consumer.
- We keep the audit gate honest by pinning the patched version rather than suppressing the advisory (`bun audit --ignore` is off the table here).

Revisit trigger + reasoning recorded in a local `dev/docs/` note: remove the override once a consumer adopts `brace-expansion^5`, `@opentui/solid` drops the build dep, or Bun offers a finer resolution. (PR description updated to state this trade-off plainly.)

PR #96, opened 2026-07-25
chore(release): sync v1.0.0-alpha.5 to develop

3aKHP:
## Summary

- forward-sync the accepted v1.0.0-alpha.5 main baseline into develop
- preserve the released Harness 10.1.2 asset baseline and post-review directory-overwrite fix

## Verification

- release workflow 30142147234 passed
- public GitHub Release checksums verified
- npm clean-install and Linux portable release-shape smoke passed


3aKHP:
Review follow-up:

- Changelog note: acknowledged as Low. This PR is an exact forward-sync of the already-published main baseline, so adding a develop-only changelog edit here would create branch drift. The directory-target fix remains documented in PR #94 and its resolved review thread; no release-blocking ambiguity remains.
- Harness hashes: independently confirmed against the official NN harness-20260724-2 ZIP and its published SHA256 before bundling. The public Vesicle assets ZIP was then downloaded afresh from Release v1.0.0-alpha.5, verified against SHA256SUMS.txt, and unzip-tested; its harness-manifest.json is therefore the exact released content.

All forward-sync CI checks are green after rerunning the Windows cleanup-timing flake (initial EBUSY occurred only while deleting the completed smoke temp directory).

PR #103, opened 2026-07-25
fix(tui): keep light-theme code blocks readable

3aKHP:
## Summary

- pass the active Vesicle foreground to OpenTUI Markdown so unlabelled fenced code does not fall back to white
- refresh mounted Markdown renderables when `/theme` or terminal auto-detection changes the resolved theme
- cover both light-theme initial mount and the production-preloaded dark-to-light transition

## Root cause

`MarkdownContent` supplied a `SyntaxStyle` but no base `fg`. OpenTUI's unlabelled fenced-code fallback therefore used `CodeRenderable`'s white default. In addition, the mounted `MarkdownRenderable` retained its initial foreground and syntax style across a live theme change unless its styles were explicitly refreshed.

## Verification

- `bun run lint`
- `bun run typecheck`
- `bun test`
- `bun run doctor`
- `git diff --check`


3aKHP:
Addressed the review findings in f2c4405: the production-preloaded probe now renders MarkdownContent through JSX, the subprocess has a bounded timeout, the colour-span helper is shared, and the component seeds its applied style state so initial mount does not call refreshStyles redundantly. Focused tests, lint, typecheck, full bun test (1080 pass / 5 platform skips), and diff check all pass locally.

PR #114, opened 2026-07-26
fix(compact): harden automatic compaction boundaries

3aKHP:
## Summary

- harden provider-send budget enforcement across pre-turn, mid-turn, resumed continuation, queued input, and background notification boundaries
- make compact checkpoints fail closed across validation, expected-head installation, cross-process append locking, repeated retained-tail projection, and unresolved interactions
- reject replacement histories that do not reduce the next request, preventing repeated non-beneficial checkpoint churn discovered during real Terminal dogfood
- align `/context`, provider configuration validation, lifecycle events, TUI draft handling, and paired user/developer documentation

## Supplemental CR Scope

This PR follows up the merged implementation from #109, #110, #112, and #113. Parallel review covered budget/provider-send boundaries, session/checkpoint integrity, and TUI/lifecycle observability. Per maintainer direction, no additional independent CR is required after this PR; Bot Review findings will be handled before merge.

## Real Terminal Dogfood

Used the dedicated `dev/test-config` provider credentials through an interactive 80-column TUI. The native MiniMax streaming endpoint reproduced `Provider stream ended before [DONE]`, so a temporary local non-stream compatibility proxy was used for the compaction workflow. Dogfood exposed repeated checkpoints whose replacement estimate exceeded the pre-compact projection; the new non-reduction guard prevents that churn while allowing the original request to continue below the hard ceiling. Temporary provider state, proxy, projects, and credentials are not included in this branch.

## Verification

- `bun run lint`
- `bun run typecheck`
- `bun test` (`1248 pass`, `5 skip`, `0 fail`)
- `bun run doctor` (`Missing: none`)
- `git diff --check`
- focused pre-turn and mid-turn auto-compaction integration suites (`9 pass`, `0 fail`)


3aKHP:
@KHPilot[bot] Please review the current head.

3aKHP:
## Bot Review disposition

Current head: `fe904b1`

- Fixed the cleanup finding: post-append file-handle and lock-file cleanup is best-effort and can no longer mask a durable successful append.
- Fixed the attachment-path finding: `.vesicle/attachments/..` now fails checkpoint validation, with focused regression coverage.
- No change for malformed prior checkpoints: `compact-checkpoint-v1` is durable replacement authority, and `docs/dev/STYLE.md` requires malformed v1 payloads to fail with an actionable session error rather than be partially projected. Continuing without reconstructing the retained replacement tail would violate that contract; the normal session projector rejects the same payload as well.
- No change for locking only conditional appends: `appendIfHead` is atomic across processes only when every competing ordinary append observes the same file lock. Allowing `appendMany` to bypass it would reopen the head-read/append race the lock fixes.
- No change for zero-growth `usageSource`: when estimated growth is zero, the projected token value is exactly the provider-observed occupancy. The source label records which input determines the projected value; marking that value `estimated` would be less accurate. Positive host-estimated growth remains labeled `estimated`.

Verification: `bun run lint`, `bun run typecheck`, `bun test` (`1248 pass`, `5 skip`, `0 fail`), `bun run doctor`, and `git diff --check`.

PR #145, opened 2026-07-30
feat(providers): add Responses WebSocket lifecycle

3aKHP:
## Summary

- add opt-in session-scoped WebSocket transport for the independent OpenAI Responses adapter
- share the existing deterministic request codec, typed event accumulator, terminal parser, native state, and commit barrier with HTTP/SSE
- implement public and frozen Codex-beta request profiles, `generate: false` prewarm, incremental `previous_response_id` continuation, full-context recovery, five-retry policy, and permanent session HTTP downgrade
- enforce one in-flight response, owner/session/process cleanup, connection rotation, and fail-closed profile/transport validation

Closes the WebSocket lifecycle phase of #122. Remote compaction, third-party subset exposure, and final product documentation remain separate follow-up phases.

## Verification

- `bun run lint`
- `bun run typecheck`
- `bun test` (full repository suite)
- `bun run doctor`
- `bun run pack:check`
- `bun run pack:smoke` (global/local package install, audit, PTY TUI startup)
- `BUN_E2E_RESPONSES_PROVIDER=doro-gpt bun run test:acceptance:responses` (real HTTP/SSE function loop and native Item replay: pass)
- direct sanitized `doro-gpt` WebSocket capability probe (handshake unavailable; not claimed as WebSocket evidence)
- `git diff --check`

## Evidence boundary

Current OpenAI WebSocket documentation was rechecked for request shape, prewarm, continuation recovery, one-in-flight behavior, and the 60-minute connection limit. Official OpenAI credentials were unavailable, so the PR does not claim a live official-endpoint pass. The real `doro-gpt` run proves the HTTP fallback endpoint only; its WebSocket handshake was unavailable.


3aKHP:
Addressed review findings in `a46a7a4`:

- Bun 1.3.14 explicitly supports constructor `headers` as a native WebSocket extension. The `as never` workaround is replaced by an explicit `Bun.WebSocketOptions` constructor overload, and a real local Bun upgrade server now asserts that `Authorization` reaches the handshake.
- Pending opens now have deterministic cancellation/rejection, reject stale late-open callbacks, and actively close on open error. A real child-process test holds a native socket open, sends SIGTERM, and verifies exit 143.
- `previous_response_not_found` now propagates through structured `ProviderError.code`; retry logic no longer parses error prose.
- Prewarm events remain internal and were never yielded to consumers; successful-attempt assertions continue to observe exactly one `attempt_started`. Prewarm now additionally requires terminal `output: []`.
- Independent CR found retry-budget stacking and delayed provider-switch teardown. The HTTP downgrade is now one final attempt after the six WS attempts, and provider/model configuration changes close current session resources before owner signals change.
- Added canonical SSE/WS fixture equivalence, fallback-HTTP failure budgeting, TUI selection lifecycle, open-error cleanup, and a real Agent Loop WS function-call → `write_file` → tool-result continuation on one socket.

Focused verification: 39 tests passed before the expanded full repository suite; full `bun test`, lint, typecheck, doctor, and `git diff --check` are green locally. Current exact head CI is running.

3aKHP:
@KHPilot[bot] Please re-review exact head `a46a7a4`, including the evidence and fixes in the preceding comment. Report any remaining Blocking or Should-fix findings.

3aKHP:
Addressed the remaining independent Should-fix finding in `f1cfd0f`:

- removed all process signal handlers and `process.exit()` calls from the provider transport;
- added a host-owned shutdown lifecycle in the CLI;
- TUI background-process and side-question cleanup now completes before provider sessions close;
- SIGINT/SIGTERM preserve exit codes 130/143 only after registered async cleanup completes;
- updated the real child-process SIGTERM test to prove host cleanup completion, provider socket release, and exit 143.

Full `bun test`, lint, typecheck, doctor, `git diff --check`, `pack:check`, and `pack:smoke` are green locally.

@KHPilot[bot] Please re-review exact head `f1cfd0f`, directly checking the remaining lifecycle ownership finding and reporting any Blocking or Should-fix issues.

3aKHP:
Follow-up exact head is now `c23ec1e`.

The previous Bot re-review's two medium points are resolved or explicitly fixed:

- An uncommitted malformed/failed WS terminal intentionally resets connection-local continuation even when non-retryable; retaining that socket would reuse state that never crossed the commit barrier. The code now documents this fail-closed invariant.
- If cancellation races another provider failure, the adapter now deterministically surfaces `AbortError`, and the regression asserts its name and reason.

The two low state-machine ambiguities are also clarified: a stale open failure no longer double-closes/rejects, and the continuation scan documents why owner mismatch at the newest native frontier must not jump backward across a provider/model switch.

Focused WebSocket/lifecycle tests, lint, typecheck, and diff-check are green. @KHPilot[bot] Please review exact head `c23ec1e` and report only remaining Blocking or Should-fix findings.

3aKHP:
The final shutdown failure-path fix is now at exact head `cfd397e`.

A failing lower-priority host cleanup is isolated so all later cleanup owners still run. The real SIGTERM child test now deliberately rejects an async host cleanup, then proves provider cleanup still executes before exit 143. TUI side-question disposal is likewise in `finally` around background-process shutdown.

Focused tests, lint, typecheck, and diff-check are green. @KHPilot[bot] Please re-review exact head `cfd397e` for remaining Blocking or Should-fix findings.

3aKHP:
Final candidate head is `6caee8e`.

Addressed the latest exact-head Bot findings:

- each WS response wait has a 120-second terminal-event timeout; focused tests inject 5 ms and prove the socket is closed with a retryable structured error;
- a session permanently disabled after WS exhaustion now reports `retryable: false`;
- removed the ineffective async `exit` hook (natural `beforeExit` and awaited SIGINT/SIGTERM paths remain);
- guarded terminal settlement and deduplicated the test reset helper;
- fixed the HTTP JSON retry path so parsing succeeds before `attempt_started` is published, preventing duplicate attempt starts on malformed JSON.

Focused 32-test provider/lifecycle suite, lint, typecheck, and diff-check are green. @KHPilot[bot] Please review exact head `6caee8e` and report only remaining Blocking or Should-fix findings.

3aKHP:
Latest exact head is `d1bf74c`.

Addressed the two remaining medium findings from the `6caee8e` review:

- TUI normal teardown and signal teardown now share the same cached async host-cleanup promise; provider sessions close only after background-process and side-question cleanup settles.
- `abortError` now has one shared provider implementation.

Also documented why `/btw` intentionally omits session ownership: it is an independent, tool-free request and must not contend with the main turn's one-in-flight WebSocket or mutate its continuation chain.

Focused 59-test suite, lint, typecheck, and diff-check are green. @KHPilot[bot] Please re-review exact head `d1bf74c` and report only remaining Blocking or Should-fix findings.

3aKHP:
@KHPilot[bot] Exact-head CI is now fully green for `d1bf74c` (source/package, Linux ELF, Windows installer). Please complete the requested exact-head re-review and report any remaining Blocking or Should-fix findings.

3aKHP:
Final exact head is now `0f572ed`.

Addressed the latest Bot medium and the related active-request edge:

- SIGINT/SIGTERM cleanup has a 10-second upper bound before forced exit, so a hung cleanup cannot trap the process indefinitely;
- session close now explicitly aborts the active WS request before closing the underlying socket, so it does not depend on a socket implementation emitting `close`;
- regression coverage proves an active request rejects with the host teardown reason and the socket is closed.

Focused 51-test provider/CLI suite, lint, typecheck, and diff-check are green. @KHPilot[bot] Please re-review exact head `0f572ed` and report only remaining Blocking or Should-fix findings.

3aKHP:
Exact-head Windows smoke failure classification for `0f572ed`:

- unrelated test: `process runtime > runs Windows windows-powershell-5.1 with UTF-8 output`;
- it hit the existing 15-second process timeout and returned exit 1;
- this PR delta does not touch Process Runtime;
- the same test passed on the preceding exact head in 4.48 seconds, while PowerShell 7, cmd, Git Bash, and auto-through-5.1 passed in this run.

Treating this as runner/timing flake, not as passed. Rerunning only the failed exact-head job; merge remains blocked until it is green.

3aKHP:
Final candidate head advanced to `902ddba`.

Addressed the valid parts of the `0f572ed` Bot review:

- snapshot `VesicleRequest.messages` once at the start of the WS logical request, so prewarm/generate/retry/fallback share one immutable message frontier;
- wrap synchronous `socket.send()` failures through the same settlement cleanup path, with a regression proving rejection and socket close;
- make the `matches()` session-id invariant explicit.

`inFlight` intentionally remains owned by the active request until its abort rejection reaches `finally`; clearing it synchronously in `close()` would permit a second request to overlap the old one. This invariant is now documented.

Focused 34-test provider suite, lint, typecheck, and diff-check are green. @KHPilot[bot] Please re-review exact head `902ddba` and report only remaining Blocking or Should-fix findings.

3aKHP:
Exact head advanced to `3a40b58` after independent CR reproduced one Should-fix.

Fixed and protected at the adapter boundary:

- synchronous `socket.send()` exceptions are now retryable provider transport failures instead of raw, non-retryable errors;
- regression proves six bounded WebSocket attempts followed by exactly one HTTP downgrade attempt;
- the same regression mutates the caller's message array after the first send and proves every WS attempt plus HTTP fallback uses only the frozen original message frontier.

Focused 22-test WebSocket suite, lint, typecheck, and diff-check are green. @KHPilot[bot] Please re-review exact head `3a40b58` and report only remaining Blocking or Should-fix findings.

3aKHP:
Exact head advanced to `a3a538e` to address the remaining valid Medium from the Bot's `902ddba` review.

Session teardown now isolates each connect-cancel, request-abort, active-socket close, and pending-socket close action, so one faulty native cleanup cannot short-circuit the rest. The pending-connect cancellation itself also rejects the waiting request even if listener removal throws. A focused regression proves listener cleanup may throw while `close()` stays non-throwing, the pending request rejects with the teardown reason, and the socket is still closed.

Focused 23-test WebSocket suite, lint, typecheck, and diff-check are green. @KHPilot[bot] Please re-review exact head `a3a538e` and report only remaining Blocking or Should-fix findings.

3aKHP:
Exact head advanced to `951a2be` to address the `a3a538e` Bot Medium.

The WebSocket logical request now deep-clones all serializable request data once while retaining only runtime `AbortSignal` and retry-callback identities. The retry/fallback regression now mutates the caller's messages, nested tool definition, and generation limit after the first send; all six WS attempts and the sole HTTP fallback continue to carry only the original values.

The pre-aborted-signal Low does not reproduce as described: listeners are registered before the final `signal.aborted` check, then `aborted()` removes them and clears `cancelConnect`; execution does not attach or overwrite them afterward. The FakeSocket `{ once: true }` fidelity note is non-blocking and does not affect the asserted single-terminal cases.

Focused 23-test WebSocket suite, lint, typecheck, and diff-check are green. @KHPilot[bot] Please re-review exact head `951a2be` and report only remaining Blocking or Should-fix findings.

3aKHP:
Exact head advanced to `d9428a3` to close both Medium findings from the exact `951a2be` Bot review.

- `receiveTerminal` now owns rejection/listener cleanup only; `session.request()` recovery owns the single socket close for both caller cancellation and host/session teardown.
- the duplicate adapter-level `prepareForRequest()` call is removed; `ResponsesWebSocketSession.request()` remains the sole pre-request rotation owner.
- regressions now require exactly one socket close for active session teardown and caller cancellation.

Focused 23-test WebSocket suite, lint, typecheck, and diff-check are green. @KHPilot[bot] Please re-review exact head `d9428a3` and report only remaining Blocking or Should-fix findings.

3aKHP:
Exact head advanced to `73d019c` to close the `d9428a3` Bot Medium.

The WebSocket handshake phase now uses the same bounded request timeout as the post-open terminal-event phase. A handshake that emits no `open`, `error`, or `close` event rejects as a retryable transport failure, clears connection state/listeners/timer, and closes the pending socket. A focused regression uses a socket that never completes its handshake and proves timed rejection plus deterministic close.

Focused 24-test WebSocket suite, lint, typecheck, and diff-check are green. @KHPilot[bot] Please re-review exact head `73d019c` and report only remaining Blocking or Should-fix findings.

3aKHP:
Exact head advanced to `ec3a409` after independent CR found one Phase 4 blocker.

Fixed the noninteractive lifecycle deadlock:

- `vesicle once` now awaits host cleanup in `finally`, because an open session socket can itself prevent `beforeExit`;
- a native-socket child-process regression completes a response, explicitly runs the same cleanup registry, proves the server observes socket close, and proves the process exits 0 without a signal.

The review's second finding—session ownership for child-Agent and compaction provider calls—is valid but is explicitly frozen Phase 5 scope (`child-Agent scenario coverage` plus dual compaction lifecycle). It will be the first Phase 5 lifecycle regression rather than being folded into this Phase 4 PR.

Focused 25-test WebSocket suite, lint, typecheck, and diff-check are green. @KHPilot[bot] Please re-review exact head `ec3a409` and report only remaining Blocking or Should-fix findings.

3aKHP:
Exact head advanced to `7fae1f6` to close the valid Phase 4 finding from independent exact-head CR.

A completed response now arms a session-owned idle close/error listener. The next request disarms it before becoming active; an idle disconnect clears the socket, response ID, and prewarm eligibility before any later request can reuse connection-local continuation on a new socket. A focused regression completes a response, closes the idle peer, proves continuation is cleared, and proves the next request creates a fresh socket.

The independent CR's compaction-invalidation finding remains the first mandatory Phase 5 checkpoint/continuation regression, matching the frozen phase boundary.

Focused 26-test WebSocket suite, lint, typecheck, and diff-check are green. @KHPilot[bot] Please re-review exact head `7fae1f6` and report only remaining Blocking or Should-fix findings.

3aKHP:
Exact head advanced to `ce7f441` to close the valid findings from the exact `7fae1f6` Bot review.

- Host shutdown is now an explicit one-shot transaction: the settled promise prevents duplicate resource cleanup, and registration after shutdown starts throws instead of being silently skipped. The noninteractive native-socket regression proves the late-registration rejection as well as clean process exit.
- `markCompleted(undefined)` now fails safe by clearing continuation/prewarm state.

The alleged idle-listener leak does not occur: `armIdleLifecycle()` calls `disarmIdleLifecycle()`, whose stored cleanup explicitly removes both prior `close` and `error` listeners with `removeEventListener` before new listeners are registered.

Focused 26-test WebSocket suite, lint, typecheck, and diff-check are green. @KHPilot[bot] Please re-review exact head `ce7f441` and report only remaining Blocking or Should-fix findings.

3aKHP:
Exact head advanced to `e413b98` after independent review found a synchronous re-entrancy window in the one-shot shutdown guard.

`shutdownStarted` is now set synchronously before any cleanup callback can run, so both post-settlement and in-cleanup registration fail explicitly. The native noninteractive child regression proves both rejection paths while still closing the socket and exiting 0.

Focused 26-test WebSocket suite, lint, typecheck, and diff-check are green. @KHPilot[bot] Please re-review exact head `e413b98` and report only remaining Blocking or Should-fix findings.

3aKHP:
Exact head advanced to `6db0aa1` to close the sole Medium from the exact `e413b98` Bot review.

Injected socket-factory identity now participates in session matching, with a regression proving same session/owner plus a different factory creates a new session. Default production factory identity remains stable (`undefined === undefined`).

The pre-aborted-signal Low is not applicable: `AbortSignal.addEventListener` does not replay a past abort event, so the explicit `signal.aborted` check is required. The future hypothetical non-cloneable request field is not a current contract defect.

Focused 26-test WebSocket suite, lint, typecheck, and diff-check are green. @KHPilot[bot] Please re-review exact head `6db0aa1` and report only remaining Blocking or Should-fix findings.

3aKHP:
Exact head advanced to `3cb6c1d` to close the valid findings from the exact `6db0aa1` Bot review.

- Non-cloneable request data now fails before socket creation with a structured, provider-qualified malformed-response error; regression uses a callable metadata value and proves zero sockets open.
- Idle listener installation now rechecks `readyState` after listeners and cleanup ownership are installed, immediately invalidating a socket that closed during the transition.

Interactive teardown does not depend on `beforeExit`: the TUI `onCleanup` explicitly closes provider sessions, while SIGINT/SIGTERM own process-signal teardown. `beforeExit` remains only a no-handle fallback.

Focused 27-test WebSocket suite, lint, typecheck, and diff-check are green. @KHPilot[bot] Please re-review exact head `3cb6c1d` and report only remaining Blocking or Should-fix findings.

3aKHP:
Exact head advanced to `932a579` to close both cleanup Medium findings from the exact `3cb6c1d` Bot review.

Every listener removal in connect, terminal receive, and idle teardown is now individually best-effort, while timeout clearing always executes. The injected listener-removal failure regression now waits beyond the configured handshake timeout and proves no delayed second close/action occurs.

Focused 27-test WebSocket suite, lint, typecheck, and diff-check are green. @KHPilot[bot] Please re-review exact head `932a579` and report only remaining Blocking or Should-fix findings.

3aKHP:
Exact head advanced to `b4d7f04` to close the valid continuation TOCTOU finding from the exact `932a579` Bot review.

`ResponsesWebSocketSession.request()` now revalidates the expected response ID after connect and immediately before send. If idle invalidation cleared the chain, no stale frame is sent and the existing retry/prewarm recovery takes over. A focused race regression closes the idle socket after request start but before its continuation microtask dispatch, then proves send count remains unchanged.

The alleged duplicate HTTP `attempt_started` is not externally observable: `readResponsesStream` events are buffered in `committed[]`; on stream failure that array is discarded, and the catch emits the caller's only attempt/discard pair.

Focused 28-test WebSocket suite, lint, typecheck, and diff-check are green. @KHPilot[bot] Please re-review exact head `b4d7f04` and report only remaining Blocking or Should-fix findings.

3aKHP:
@KHPilot[bot] Please review the current exact head `5574a8b`.

The prior `response.failed` finding was addressed without fabricating an HTTP status: the adapter now preserves the provider error code, retries error-bearing failures except the frozen profile's explicit fatal codes, and has focused server-error recovery plus fatal context-length coverage. Official public WebSocket docs explicitly say streaming is implicit, so `stream` remains omitted for `openai-public`. The shutdown cleanup transaction catches each registered cleanup internally and is not expected to reject.

3aKHP:
@KHPilot[bot] Please review the current exact head `f2a328f`.

The independent incremental CR requested table-driven coverage for every frozen fatal `response.failed` code; the focused regression now covers all six while retaining the retryable server-error recovery case.

PR #181, opened 2026-08-07
fix(checkpoint): exclude scratch tmp/ from rewind checkpoints (#137B slice 0)

3aKHP:
## Summary

- **Slice 0 of #137B** (independent prerequisite PR). Design freeze: [#137 (comment)](https://github.com/3aKHP/prism-vesicle/issues/137#issuecomment-5222773633).
- `trackBeforeMutation` now skips scratch paths via the existing `isScratchProjectPath`, so `tmp/` mutations are **writable but no longer tracked by the per-turn checkpoint/rewind ledger**. This reverses the 137A decision and prepares `tmp/` to hold ephemeral MCP tool-result spills without polluting the durable content-root ledger.
- Write permission is unchanged (`modelWritableRoots` still includes `tmp/`); only the checkpoint scope narrows. All six `trackBeforeMutation` callers flow through this one method.
- Inverts the scratch checkpoint test into a selectivity boundary test: in the same turn a `workspace/` write **is** tracked and rewound, while `tmp/` writes succeed but are **excluded**.
- Doc sweep: `STATUS` (2 spots), `README`, `docs/dev/SESSIONS.md`, `docs/dev/TOOLS.md`, the paired `permissions-and-security` user-manual pages (en + zh-CN), `CHANGELOG` `[Unreleased] › Changed`, and the auto-synced `vesicle-docs` references.

## Test Plan

- [x] `bun run lint`
- [x] `bun run typecheck`
- [x] `bun test` — 1914 pass / 0 fail (one unrelated flaky TUI component test, `compact-hero-transition`, passed on rerun; it touches no checkpoint code)
- [x] `bun run doctor` — `Missing: none`
- [x] Focused: `tests/integration/session/file-checkpoint.test.ts` (8/8), `rewind-service` + `checkpoint-ownership` + `compact-checkpoint` (16/16)

## Notes / Follow-ups

- Side effect of the reversal: `tmp/skillify/<name>/` drafts (#127) are no longer rewind-protected — acceptable for drafts; documented in CHANGELOG.
- Subsequent #137B slices (MCP persistence skeleton; auto-truncate sub-toggle) build on this and are tracked in the local working plan.
- Independent CR requested; will address Blocking/Should-fix before merge.


3aKHP:
## Independent CR — results

An independent SubAgent CR (did not participate in implementation) reviewed this branch against `develop`.

- **Blocking**: 0
- **Should-fix**: 1 → addressed in `27b0a50`
- **Nits**: 0

### Should-fix (addressed)

`createSnapshot` and `checkpointRestoreState` carried forward `tmp/` entries from 137A-era snapshots, so resuming an old session could still rewind scratch on restore — contradicting the new "scratch is not rewind-safe" contract. Fixed by filtering `isScratchProjectPath` in `checkpointRestoreState` (restore/diff safety net) and in `createSnapshot`'s prior re-capture loop (stops carry-forward + avoids redundant backup I/O). Added a cross-version resume regression test asserting a legacy snapshot's `tmp/` entry is ignored while its `workspace/` entry is honored.

### Verified claims

- Write permission unchanged: `writableFileRoots` still includes `tmp/`; `resolveAllowedPath` untouched.
- `trackBeforeMutation` is the sole entry point for new paths (all five callers route through it).
- `isScratchProjectPath` is the correct, fail-closed guard (handles bare `tmp`, `tmp/foo`, rejects non-scratch roots and path tricks).
- Inverted boundary test is non-tautological (asserts observable file contents).
- Docs fully consistent — no stale "participates in checkpoint" wording remains; the 137A CHANGELOG entry under `[1.0.0-alpha.8]` correctly stays as historical record.

### Verification (post-fix)

- `bun test`: 1915 pass / 0 fail
- `bun run doctor`: `Missing: none`
- `bun run skills:docs:check`: 83 references in sync

Merge-ready pending human approval.

3aKHP:
## Bot review — resolution

**Verdict: ✅ Approve, 0 Blocking.** Both findings addressed.

### NIT — already resolved
`createSnapshot` scratch filter: resolved in `27b0a50` (the independent-CR Should-fix commit added `isScratchProjectPath` filters to both `createSnapshot` and `checkpointRestoreState`, plus a cross-version resume regression test). This NIT was filed against `d552f18`, before that fix landed.

### SHOULD-FIX — addressed in `0e73429`
Cross-root moves across the `tmp/` boundary are asymmetric under rewind (scratch side is not checkpointed):
- **scratch → content**: promoted copy deleted on rewind, source not restored → moved body lost.
- **content → scratch**: original restored from backup, scratch copy left in place → duplicated.

Consistent with the "tmp/ is not rewind-safe" contract (tmp/ is ephemeral by design), the boundary is **documented and tested** rather than re-tracking scratch (which would re-introduce the 137A mistake). `0e73429` adds both-direction move+rewind regression tests and a caveat to SESSIONS, the paired permissions-and-security pages, STATUS, and CHANGELOG; vesicle-docs references re-synced. Workaround documented for users: use `copy_file` to promote scratch when rewind-safety matters.

### Verification (latest commit `0e73429`)
- `bun test`: 1917 pass / 0 fail
- `bun run skills:docs:check`: 83 references in sync

Proceeding to merge.

PR #182, opened 2026-08-07
feat(mcp): opt-in MCP tool-output persistence (#137B slice 1)

3aKHP:
## Summary

- **Slice 1 of #137B.** Design freeze: [#137 (comment)](https://github.com/3aKHP/prism-vesicle/issues/137#issuecomment-5222773633). Builds on slice 0 (#181, merged).
- Adds an opt-in project toggle `mcpOutputPersistence` in `.vesicle/preferences.yaml` (v1, optional, defaults `false`). When on, **every** MCP tool call's text result and decoded images are persisted under `tmp/mcp-output/<sessionId>/` (text) and `.../blob/` (images, native format) with tool/argument-derived filenames.
- **Inline delivery is unchanged** — persistence is an additive, best-effort durable copy the model re-reads with existing `read_file` / `grep_files` / `view_image` tools instead of repeating an expensive or non-repeatable MCP call. A persistence failure never alters the tool result.
- **No new tool, no MCP schema injection.** The toggle + sessionId ride `McpRegistryOptions` (registry captures them at build), so `tool-round-executor` is untouched. The hint is appended by the prompt builders and re-appended each round by `refreshLiveSystemPrompt` (mirroring the Skill catalog block), gated on the toggle **and** actual MCP availability (`mcpRegistry.definitions.length > 0`).
- `writeProjectThemePreference` round-trips `mcpOutputPersistence` (writing theme does not wipe the toggle); malformed/absent preferences read as off.
- Auto-truncation and bounded `read_file` are deferred to slice 2.

## Test Plan

- [x] `bun run lint`
- [x] `bun run typecheck`
- [x] `bun test` — 1926 pass / 0 fail
- [x] New: `tests/integration/tools/mcp-output-persistence.test.ts` — helper writes text + native images with meaningful names; `deliverMcpToolResult` persists when `outputPersistence` set and writes nothing when absent; hint advertises path + readback tools.
- [x] New: `tests/unit/config/project-preferences.test.ts` — toggle off by default, on only when `true`, malformed reads as off, theme write preserves the toggle.

## Notes / Follow-ups

- Slice 2: auto-truncate sub-toggle + length threshold + `read_file` bounded byte-offset read + `grep_files` excerpt cap.
- Independent SubAgent CR requested; will also address KHPilot Bot Review findings before merge.


3aKHP:
## Independent CR + Bot Review — resolution

Both reviews completed. **0 Blocking from either.** All Should-fix and the substantive Nits are addressed in `b011cf6`.

### Independent SubAgent CR — 0 Blocking, 3 Should-fix (fixed), 2 Nits
1. **`unsetProjectThemePreference` wiped `mcpOutputPersistence`** when removing the theme (the symmetric case of the write round-trip) → fixed: now rewrites the file preserving the toggle; only unlinks when no field remains. Regression test added.
2. **No path-traversal test** → added an adversarial test (`../../etc`, `/absolute/path`, `..\..\windows`, `a/b/c`, NUL) proving nothing escapes `tmp/mcp-output/<sessionId>/` and no filename contains `..`/`/`/`\`.
3. **Misleading `persistMcpOutput` docstring** ("Never throws") → reworded to "May throw … caller treats as best-effort."
- Nits: reworded the `composeMcpOutputPersistenceHint` docstring; left `sessionId` (host-generated, safe) and the abort guard (harmless defense-in-depth for the no-image path) as-is per reviewer.

**Path-traversal verdict (CR): SUFFICIENT.** `sanitizeFileName` collapses every vector (`..`, `/`, `\`, NUL, case/Unicode) to `[a-z0-9_-]`; `sessionId` is host-generated. Writes follow the attachment-store precedent (host-internal side effect, not a model tool call). **Inline-result invariant: HOLDS** — persistence is additive, best-effort, `.catch`-guarded, never touches the returned content. **Wiring: all 6 RunLoopArgs sites wired; session-init correctly skipped; hint gated on toggle + `definitions.length > 0`.**

### KHPilot Bot Review — 💬 Comment (no issues) + Style ✅ Approve, 2 Nits
Both Nits (the same two docstrings) were already resolved in `b011cf6`, filed against the prior commit `43338db`.

### Verification (latest `b011cf6`)
- `bun test`: 1928 pass / 0 fail
- `bun run skills:docs:check`: 83 in sync

Proceeding to merge once CI is green.

PR #183, opened 2026-08-08
feat(mcp): MCP output auto-truncate + bounded read_file + grep cap (#137B slice 2)

3aKHP:
## Summary

- **Slice 2 of #137B (final).** Design freeze: [#137 (comment)](https://github.com/3aKHP/prism-vesicle/issues/137#issuecomment-5222773633). Builds on slice 0 (#181) and slice 1 (#182).
- **Auto-truncate sub-toggle:** `mcpOutputAutoTruncate: true` in `.vesicle/preferences.yaml` (requires `mcpOutputPersistence`). When on, an MCP text result at or above **32 KiB** is delivered inline as a **4 KiB preview + reference** to the persisted full copy, so a single large result cannot dominate the next provider request. Below the threshold the full body stays inline; either way the complete text is on disk.
- **Bounded `read_file`:** optional `offsetBytes`/`maxBytes` read a bounded UTF-8 byte slice **without loading the whole file** (project files only) — for large persisted outputs and giant single-line payloads. Existing line-range behavior unchanged when `maxBytes` is absent.
- **`grep_files` excerpt cap:** each match-line excerpt is capped at 500 characters so one giant (e.g. minified) line can't fill the result.
- `preferencesBody` now round-trips both toggles (`/theme --persist` and unset preserve them). `readMcpOutputPreferences` reports `autoTruncate` as on only when persistence is also on. The flag rides `McpRegistryOptions` alongside the persistence toggle.
- Byte-based thresholds keep CJK and English bounded consistently.

## Test Plan

- [x] `bun run lint`, `bun run typecheck`: clean
- [x] `bun test` — 1951 pass / 0 fail
- [x] New: truncation over/under threshold + preview-contains-reference + persisted-full-copy; `readMcpOutputPreferences` gating (autoTruncate off unless persist on); malformed value diagnostic; theme-write preserves both toggles.
- [x] New: `read_file` bounded slice (offset/maxBytes/from-start/full-not-truncated/assets-rejected); `grep_files` giant-line excerpt cap.
- [x] Re-tuned 3 auto-compaction tests (minimal `defaultTools:[read_file]` surface) to wider `contextWindow` margins — the `read_file` schema growth had exceeded their ~50–64 token hard-ceiling margins.

## Notes / Follow-ups

- The 32 KiB / 4 KiB constants are tunable defaults (documented in `output-persistence.ts`), pending broader fixture validation; not user-configurable (the sub-toggle is the user knob).
- Independent SubAgent CR requested; will also address KHPilot Bot Review findings before merge.


3aKHP:
## Independent CR + Bot Review — resolution

Both reviews complete. **0 Blocking from either.** All Should-fix and the substantive nits are addressed.

### Independent SubAgent CR — 0 Blocking, 1 Should-fix (fixed), verified the re-tune
- **Should-fix (fixed `2e29127`):** `read_file` with `offsetBytes` but no `maxBytes` silently fell through to the full-file path. Now throws `offsetBytes requires maxBytes`; regression test added.
- **Verified:** inline-result invariant holds (off / under-threshold byte-identical; persistence failure cannot truncate); `readByteSlice` never loads the whole file, handles EOF/empty/clamping; grep cap preserves match semantics; config round-trip + `autoTruncate`-requires-`persist` gating; threshold rationale sound.
- **Verified the compaction-test re-tune is legitimate:** computed the actual read_file schema growth (+643 JSON bytes ≈ +322 estimated tokens) and confirmed it exceeded the old tests' 29–211 token margins (old Turn-1/pre-turn estimates would have crossed the threshold). New margins are 530–2258 tokens, stable; all behavioral assertions preserved; no coverage weakened.

### KHPilot Bot Review — ✅ Approve, 0 Blocking, 1 Should-fix + nits
- **Should-fix (same offsetBytes issue)** — already fixed in `2e29127` (filed against the prior commit).
- **Nits (addressed `a4af7d4`):** `readByteSlice` now uses actual `bytesRead` (no NUL injection / over-report on a short read); stale module header qualified; dropped the unused `imagePaths` return surface; `readMcpOutputPersistence` reduced to a thin wrapper.

### Verification (latest `a4af7d4`)
- `bun test`: 1952 pass / 0 fail
- `bun run skills:docs:check`: 83 in sync

Proceeding to merge once CI is green.

PR #185, opened 2026-08-08
fix(skills): host-inject Skill tools, decoupling from Harness profiles

3aKHP:
## Summary

- **Root cause**: The Harness 10.2.0 bump (`f964c41`) replaced `assets/engines/` wholesale and silently dropped `activate_skill`/`read_skill_resource`/`run_skill_script` from every non-Stage engine profile's `defaultTools`. The runtime gated the entire Skill surface on `profile.defaultTools.includes("activate_skill")`, so real sessions lost the `<skill_catalog>` prompt block, all Skill tools, and existing activations were pruned. Integration tests didn't catch it because their fixture wrote a synthetic profile that re-added the tools.

- **Structural fix**: Skill tools are now injected by the host layer (`resolveBuiltInTools`) for every non-Stage engine when the session catalog is non-empty, independent of the Harness profile's `defaultTools`. `resolveEngineEligibleCatalog` now gates only on `profile.id === "stage"`. Harness bumps can no longer clobber the Skill surface.

- Adds `VESICLE_HOST_ASSETS_DIR` env var override for host-assets resolution (legitimate for custom installs; used by tests to isolate from bundled Skills).

### Files changed

| File | Change |
|------|--------|
| `src/core/agent-loop/tool-surface.ts` | Move Skill tools from `defaultTools` loop to host-injection block after the loop |
| `src/core/skills/catalog-context.ts` | Simplify `resolveEngineEligibleCatalog` to Stage-only exclusion |
| `src/core/skills/catalog-sources.ts` | Add `VESICLE_HOST_ASSETS_DIR` env var support |
| `tests/unit/skills-runtime/surface-policy.test.ts` | Update 2 tests: verify injection without `defaultTools`; fix `activate_skill` enum lookup |
| `tests/unit/skills-runtime/catalog-context.test.ts` | Update eligibility test: all non-Stage engines get full catalog |
| `tests/integration/agent-loop/skill-catalog.test.ts` | Reverse "without skill tools" test: now verifies host-injection works |
| `tests/integration/agent-loop/fixtures/agent-loop.ts` | Suppress host Skills when `skillTools` is off |
| `tests/integration/agent-loop/{pre,mid}-turn-auto-compact.test.ts` | Suppress host Skills for clean compaction thresholds |
| `docs/dev/SKILLS.md` + vesicle-docs sync | Update tool registration description |
| `STATUS.md`, `CHANGELOG.md` | Document the fix |

## Test Plan

- [x] `bun run lint`
- [x] `bun run typecheck`
- [x] `bun test` (1946 pass, 0 fail)
- [x] `bun run skills:docs:check` (83 references in sync)

## Notes / Follow-ups

- The `skillTools` fixture flag is now a semantic toggle: when true, tests see real host Skills; when false, host Skills are suppressed. The flag no longer affects `defaultTools` (it's a no-op there) — it controls `VESICLE_HOST_ASSETS_DIR`.
- This does **not** add Skill tool entries back to the bundled profiles. The structural fix makes them unnecessary.


3aKHP:
## Independent CR Results

**Verdict: 0 Blocking, 0 remaining Should-fix, 2 Nits.**

### Should-fix (addressed in d28554e)

1. **Stale file-level doc comment** in `catalog-context.ts` lines 15-19 still said "an engine must also declare the Skill tools" — contradicting the function-level comment, CHANGELOG, and STATUS.md. Fixed to "Stage is the only Skill-less engine."

### Nits (deferred)

- `VESICLE_HOST_ASSETS_DIR` reads from `process.env` rather than the `env` parameter threaded through `resolveFilesystemSkills`. Not a live bug (all current callers use `process.env`), but an architectural inconsistency.
- `VESICLE_HOST_ASSETS_DIR` is undocumented in user-facing docs. Intentional as primarily a test mechanism; may warrant a brief mention later.

### Verified claims

- Injection logic correct: non-Stage + non-empty catalog → `activate_skill`/`read_skill_resource` always injected; `run_skill_script` additionally gated on process capability
- No duplicate-tool risk: `defaultTools` loop explicitly skips all three skill tool names
- `byName.get(...)!` non-null assertions are safe (definitions guaranteed in static pool or dynamically registered)
- `resolveEngineEligibleCatalog` simplification safe across all call sites; no caller relied on the old buggy behavior
- `VESICLE_HOST_ASSETS_DIR` has no security implications (same path hardening applies regardless of root)
- Test env cleanup correct in all files; no env var leaks between tests
- Regression coverage sufficient: unit test proves injection without `defaultTools`; integration test proves end-to-end; Stage exclusion verified

PR #186, opened 2026-08-08
feat(tools): grep_files output modes, context lines, and output budget

3aKHP:
## Summary

- **`outputMode`** parameter: `content` (default, unchanged behavior) | `files_with_matches` (returns only file paths that contain matches, early-exit per file) | `count` (per-file match counts + total). `maxMatches` limits match entries in content mode and file count in the other two.
- **`contextLines`** parameter (0–10, default 0): in content mode, each match entry includes `before`/`after` arrays of `{line, text}` context lines, capped at 500 chars per line via existing `capGrepExcerpt`. Eliminates the common grep → read_file follow-up round-trip.
- **Output budget**: host-side 32 KiB text ceiling (`maxGrepResultChars`) in content mode, acting as a safety valve alongside `maxMatches` (50 default, 200 cap).
- **Shared scanning core**: extracted `executeGrepScan` to avoid 6× duplication between `grepFiles`/`grepAssetFiles` across three modes. `normalizeLines` strips the trailing empty element from a terminal newline so it does not appear as a spurious context line.
- **TUI footer**: `files_with_matches` mode shows file count instead of match count.

### Files changed

| File | Change |
|------|--------|
| `src/core/tools/file/definitions.ts` | grep schema: add `contextLines` + `outputMode`; update description |
| `src/core/tools/file/query-operations.ts` | core: `executeGrepScan` shared scan, three mode branches, context collection, budget, `normalizeLines` |
| `src/core/tools/file/read-handlers.ts` | handler: parse new params, mode-aware `FileToolEvent` |
| `src/core/tools/types.ts` | `FileToolEvent`: add `outputMode?` + `fileCount?` |
| `src/tui/tool-render.ts` | grep footer: `files_with_matches` branch |
| `tests/integration/tools/file-tools.test.ts` | 6 new tests (context basic/boundary/cap + 3 output modes) |
| `docs/dev/TOOLS.md` + `dev-tools.md` | grep output mode / context / budget documentation |
| `CHANGELOG.md` | `[Unreleased] > Added` entry |

### Downstream safety

Verified that grep `FileToolEvent` is opaque everywhere except `tool-render.ts` (null-guarded `matches`/`truncated`). Checkpoints, session, quality, and compact all pass it through without field inspection. The content JSON shape is consumed only by the LLM.

## Test Plan

- [x] `bun run lint` — 682 files clean
- [x] `bun run typecheck` — no errors
- [x] `bun test` — 1952 pass, 0 fail (includes 6 new grep tests)
- [x] `bun run doctor` — no issues

## Notes / Follow-ups

- Glob filtering (`include`/`exclude`) is deferred to a separate dedicated glob tool evaluation.
- Regex engine remains trusted single-user (no RE2/worker isolation) — pre-existing, unchanged.


3aKHP:
## CR Summary

Both reviews returned **✅ Approve** with no blocking findings. All should-fix items have been addressed in the latest push (e3c87aa).

### SubAgent CR — Should-fix → Fixed

| # | Finding | Fix |
|---|---------|-----|
| 1 | `outputMode` had no runtime validation — `"bogus"` silently fell through to content mode | Added `validateOutputMode()` guard in `query-operations.ts`; rejects unknown values with a clear error |
| 2 | 32 KiB output-text ceiling was untested | New test: 200×300-char matching lines with `maxMatches:200` → asserts `truncated:true` and `matches.length < 200` |
| 3 | `count` mode `maxMatches` truncation untested | New test: 3 files, `maxMatches:2` → asserts `counts.length === 2` and `truncated:true` |
| 4 | TUI `files_with_matches` footer untested | New `toolResultFooter` unit test: `outputMode:"files_with_matches"`, `fileCount:5` → `"5 files"` |

### Bot CR — Should-fix → Fixed

| # | Finding | Fix |
|---|---------|-----|
| 1 | Same as SubAgent #1–#3 (budget + count truncation untested) | Same fixes |
| 2 | `GrepOutputMode` exported but never imported; inline copies in `read-handlers.ts` and `types.ts` | `read-handlers.ts` now imports `GrepOutputMode` from `query-operations.ts`, eliminating one of three copies. `types.ts` keeps its inline copy (importing from a specific implementation module into shared types violates dependency direction). |

### Nits (not addressed — documented for awareness)

- **`normalizeLines` behavior change**: files ending with `\n` no longer produce a phantom trailing empty line. This is a correct improvement (the phantom line was never real) but is technically an unannounced change to existing content-mode grep behavior. Only affects patterns like `^$`.
- **Context line overlap**: nearby matches independently report overlapping context (unlike `grep -C` which merges). Acceptable trade-off for simplicity; the output budget mitigates token waste.
- **Nested ternary** for `FileToolEvent` construction and **`GrepArgs` not exported**: style preferences, not correctness issues.

### Verification

```
bun run lint      ✅ 682 files clean
bun run typecheck  ✅ no errors
bun test           ✅ 1952 pass, 0 fail (61 in targeted files, +9 new tests)
bun run doctor     ✅ no issues
```

PR #213, opened 2026-08-14
feat: any-depth candidate tree with full-manifest file switching

3aKHP:
## Summary

- **Any-depth candidate switching.** Stored candidate trees were reachable only at the last turn; this PR makes every fork at every depth browsable and switchable. `/branch` (also `Ctrl+B` from both pages) opens a candidate-tree panel rendering all forks including branches inside inactive candidates, with fold/unfold navigation, inline excerpts, per-candidate file-state hints, a read-only file-preview confirm step, and per-fork regeneration (`r`). Switching executes through a shared kernel: files move before the selection marker, and the marker keys its fork point on the target leaf's owning turn so the inline `< n/m >` switcher re-arms at the switched-to depth.
- **Full-manifest file bundles (breaking).** `candidate-file-state` bundles become version-2 full manifests of everything on disk under the content roots; switching makes the disk strictly equal to the target manifest (out-of-manifest paths deleted, symlinks/special files exempt as `untracked`, `tmp/` excluded). Manual edits and MCP writes are now snapshotted on departure and switch with candidates. Version-1 partial bundles are rejected at parse and degrade conversation-only — alpha-stage breaking change, no migration path.
- **Unified turn-focus cursor.** `Alt+↑/↓` moves a transcript-wide turn-level focus cursor (every engine, wrap-around, brand-color highlight); `Ctrl+R` regenerates the focused turn when the cursor is set. `Alt+←/→` reports `Ctrl+B`/`Ctrl+R` guidance instead of dying silently on every surface, fixing the long-standing dead-key defect. The projector stamps `recordUuid` onto user/tool messages so anchors key on durable record identity.

Design decisions were recorded in the local research note before implementation (full tree incl. inactive subtrees with no tech debt; `/branch` + Ctrl+B; no compatibility for old bundles; ancestor forks expose only that turn's reply; wrap kept; no i18n; session-bloat governance deferred).

## Test plan

- [x] Full deterministic suite: 2096 pass / 0 fail; lint, typecheck, `vesicle doctor` clean
- [x] New focused tests: full-manifest capture/restore/deletion/symlink/v1-rejection (integration), candidate-tree primitives incl. marker-chained nested forks (unit), branch panel controller state machine (unit), `/branch` command (unit), Ctrl+B routing in both VT and Kitty encodings (component), turn-focus cursor wrap/rejection (component)
- [x] Validated against a real session file: v2 bundles, ownerFork markers, marker-chained continuations, vertical `/rewind` forks, and a genuine deep switch (turn-2 leaf selected while turn-3 candidates exist) all confirmed; validation surfaced and fixed a display scoping bug (455730c)
- [ ] Real-TUI smoke in a live terminal: build a two-candidate tree, `Ctrl+B` browse/switch at depth, `Alt+↑/↓` cursor + `Ctrl+R` focused regenerate, `Alt+←/→` guidance
- [ ] Independent CR pass (Blocking/Should-fix findings addressed before merge)


3aKHP:
Bot Review / Style Review findings addressed in d4888eb:

- **SHOULD-FIX (parallel tree walks)**: `listProjectTreePaths` / `captureProjectTree` now delegate to a single `walkProjectTree` visitor; `diffDiskAgainstEntries` is a thin wrapper over `diffCheckpointState` extended with an optional deletion list.
- **NIT (duplicated upward walk)**: `ownerForkOfLeaf` and `contentLeafAtOrAbove` share one predicate-parameterized `walkUpTo` helper.
- **NIT (multi-mechanism key handler)**: `handleStageMessageKey` split into `handleCandidateCycle` / `handleTurnFocusNavigation` / `handleStageToggle` / `handleLegacyStageNavigation` with documented precedence.
- **NIT (composition-root accumulation)**: turn-focus state, debounced refresh, and rejection guidance moved into `createTurnFocusController`.

Verification: full suite 2098 pass / 0 fail; lint and typecheck clean.

PR #227, opened 2026-08-19
fix(providers): keep Gemini image tool results out of function-response Contents

3aKHP:
Closes #226

## Summary

Repairs the Gemini `generateContent` history serializer so a `user` Content is strictly either `functionResponse`-only or ordinary multimodal-only, per the issue's required Content boundary:

- Tool branch pushes only the `functionResponse` part into the pending accumulator; tool-result images collect into a second accumulator (`pendingToolImages`).
- `flushToolResults()` emits up to two separate `user` Contents: the grouped `functionResponse` parts first (existing grouping semantics preserved), then an ordinary `user` Content with the `[Image #N: ...]` notice and `inlineData`.
- The post-flush merge branch (`if (flushedToolResults)`) is deleted; ordinary user messages after a tool round always serialize as their own Content. Ordering for `assistant(calls) → tool(c1, img) → tool(c2) → user(text)` is `model(calls)`, `user(fr c1, fr c2)`, `user(img)`, `user(text)` — images stay in the tool turn, functionResponses stay grouped.
- The change is request-serialization only and deterministic from `messages`; `response.ts`, `stream.ts`, and `types.ts` are untouched, and existing sessions replay through the new shape with no migration. `toolNameFromCallId` is unchanged (it scans `serialized` model `functionCall`s, independent of tool-result Content shape).

## Regression coverage

`tests/integration/providers/gemini-provider.test.ts`:

- `"serializes MCP tool-result images as a separate user Content after the function response"` replaces the old mixed-shape expectation: one functionResponse-only `user` Content, one `[Image #1]` + `inlineData`-only `user` Content (plus the leading `model` functionCall Content).
- `"never mixes functionResponse parts with ordinary multimodal parts in one user Content"` adds the boundary invariant across a real failure-mode sequence (two tool results, one with an image, plus a following gate-style user message), asserting both the exact serialization and the invariant that would have caught the 400.
- The comprehensive serialization test moves `[gate:phase resolved as confirm]` to its own trailing `user` Content.
- `tests/integration/agents/agent-tools.test.ts` (which also asserts a `functionResponse`) was verified unaffected — its tool message has no images.

Known behavioral edge, disclosed rather than asserted away: after the merge-branch removal, an empty user message following tool results now emits `{ role: "user", parts: [{ text: "" }] }` via the normal-path fallback instead of being silently swallowed. This matches the pre-existing fallback for ordinary empty user messages.

## Real-provider acceptance

[...]

PR #155, opened 2026-07-31
feat(skills): bundle skillify workflow with create-only draft publication (#127)

3aKHP:
## Summary

Closes #127. Delivers the first-party Host-bundled `skillify` Skill as one reviewed vertical slice with the minimum generic runtime support required across source, npm, Linux standalone binaries, and Windows packages.

Users can ask Vesicle to turn a proven workflow from the current conversation into a reusable Skill. `skillify` writes a complete draft under `tmp/skillify/<name>/`, validates it, and publishes create-only to `.agents/skills/<name>/` or the installed Skill Store after an explicit target decision. Drafts are retained, the current session catalog stays frozen, and a new session discovers the published Skill.

### Runtime and publication support

- Shared portable bundle inspection, hashing, and byte-exact staging seam.
- Name-level create-only Store publication under the cross-process lock, including verified recovery for crashes before either provenance or index commit.
- Exact `tmp/skillify/<name>` draft contract, symlink/special-file rejection, revalidated publication parents, sibling staging, atomic project rename, and stable bounded diagnostics.
- Non-model-visible `vesicle.skill-draft/v1` JSON CLI for validation and `project|installed` publication.
- Native `.ps1` execution through PowerShell 7 or Windows PowerShell 5.1 fallback, plus POSIX `.sh` support.
- Stable Vesicle self-invocation metadata for Skill scripts; first-party wrappers never guess from PATH or print executable paths.
- Two thin, equivalent wrappers: `publish_skill.sh` and `publish_skill.ps1`.
- Host-bundled `skillify` instructions, package boundary, public documentation, and synchronized `vesicle-docs` references.

### Commits

1. `feat(skills): add portable bundle seam and create-only Store publication`
2. `feat(skills): add guarded draft publisher and JSON CLI`
3. `feat(skills): support PowerShell scripts and stable self-invocation`
4. `feat(skills): bundle the skillify workflow`
5. `docs(skills): document scratch-first skillify and update pack boundary`
6. `fix(skills): harden skillify publication`

## Verification

- [x] Independent CR completed; all Blocking and Should-fix findings repaired.
- [x] `bun run lint`
- [x] `bun run typecheck`
- [x] `bun test` — 1800 pass, 6 platform skips, 0 fail
- [x] `bun run doctor`
- [x] `bun run skills:docs:check` — 82 references synchronized
- [x] `bun run pack:check` — 180 package files
- [x] `bun run pack:smoke` — global/local npm consumers, audit, and PTY startup
[...]

PR #228, opened 2026-08-19
test(skills): classify cross-process lock-test flakes and prove worker blocking

3aKHP:
## Summary

Hardens the flaky-class diagnostics of `cross-process index updates wait for the owner and recover after it exits` (`tests/unit/skills/store.test.ts`), the test that produced a one-off CI failure on the first run of #227.

- **Failure classification.** The observed CI failure was `Index worker exited before the holder: 0` — the sleeping lock holder had been killed by the environment (most plausibly a runner OOM-kill; the run sat in a fast window, worker exit 0, empty streams), SQLite released the cross-process lock with the holder's death, and the worker legitimately completed inside the 100ms liveness window. The test's premise had broken, not the store. The test now tracks holder liveness and reports its death as a distinct environmental failure, carrying the holder's exit code and the worker's state, so any future occurrence is classified in one line instead of requiring log forensics.
- **Oracle strengthening.** While the holder lives, the index write cannot have happened (it runs inside `withIndexLock`). The test now asserts the active index does not yet contain the install before killing the holder, proving the worker was genuinely blocked on the lock rather than merely slow to boot when the liveness check passes.

No production code changes; no behavioral change to the store. Timing structure (2s ready poll, 100ms observation) is unchanged.

## Analysis evidence

- CI first-run failure detail recovered from the Actions log: `error: Index worker exited before the holder: 0` at 138.78ms total — faster than the stable ~196ms local passing baseline, i.e. a fast window, ruling out slowness/timeout paths.
- `installSnapshot` on a fresh store necessarily passes `withIndexLock` (`BEGIN IMMEDIATE` retry, 50ms interval, 10s deadline), so a clean sub-100ms exit is only possible if the lock was already free — i.e. the holder process was gone.
- Not reproducible locally: 40 iterations under 4-way CPU load, all green; CI rerun of the same code green; no prior flake history for this test (introduced in #155).

## Test Plan

- `bun test tests/unit/skills/store.test.ts` × 8: 28 pass / 0 fail each run, per-test duration unchanged (~196ms vs baseline).
- `bun run lint` / `typecheck`: clean. `bun test`: 2124 pass / 0 fail / 10 skip.

### DIFF
--- a/tests/unit/skills/store.test.ts
+++ b/tests/unit/skills/store.test.ts
@@ -317,22 +317,25 @@ v2 body
           return exitCode;
         });
         await Bun.sleep(100);
-        if (holderExited) {
-          // The holder holds the SQLite write lock, so its death is the only way the
-          // worker can finish early. A dead holder is environmental (e.g. a runner
-          // OOM-kill), not a store defect, and must not be misread as a locking failure.
-          throw new Error(
+        // The holder holds the SQLite write lock, so its death is the only way the
+        // worker can finish early. A dead holder is environmental (e.g. a runner
+        // OOM-kill), not a store defect, and must not be misread as a locking failure.
+        const holderDeathError = async () =>
+          new Error(
             `Lock holder died while it should have been sleeping: exit ${await holderExit}; `
               + `worker ${workerExited ? `already exited with ${await workerExit}` : "still running"}`,
           );
-        }
+        if (holderExited) throw await holderDeathError();
         if (workerExited) {
           throw new Error(`Index worker exited before the holder: ${await workerExit}\n${await workerStdout}\n${await workerStderr}`);
         }
         // The index write happens inside the cross-process lock, so while the holder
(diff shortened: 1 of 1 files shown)


3aKHP:
## Disposition: Bot Review

- **SHOULD-FIX (TOCTOU between the liveness check and the index read)**: adopted in 339ddb4. The test now captures the pre-kill index state first, then re-checks holder liveness before asserting, so a holder death during the read classifies as the environmental failure instead of failing as an unexplained index mismatch. The shared classification message is factored into a local `holderDeathError` builder used at both check points.
- **NIT (comment overclaimed "genuinely blocked, not merely slow to boot")**: adopted — the comment now states only what the assertion demonstrates (the install cannot have landed during the observation window while the holder lives), since staging and rename run before `withIndexLock` and cannot be distinguished by this read.

Re-verified: `bun test tests/unit/skills/store.test.ts` × 5 (28 pass / 0 fail each), `lint` / `typecheck` clean.

PR #233, opened 2026-08-20
fix(config): preserve provider registry source edits

### LINKED ISSUE #232: fix(config): preserve providers.yaml content when update-config adds a model
## Problem

`update-config` delegates `add-model` to `vesicle config add-model`. That command loads the provider registry into a normalized in-memory model and then serializes the entire `providers.yaml` again.

This is not safe for a user-owned configuration file: an operation that should append one model silently rewrites unrelated content.

## Observed regression

During the built-in web search rollout, existing Gemini model entries had:

```yaml
capabilities:
  builtinWebSearch: true
webSearchDefault: true
```

Adding `gemini-2.5-pro` through the `update-config` Skill passed an entry that also included `webSearchDefault: true`. After the command succeeded, every `webSearchDefault` field had disappeared from `providers.yaml`.

The direct cause is `serializeModel()` in `src/setup/config-writer.ts`: it writes `generation`, `capabilities`, and `limits`, but omits `webSearchDefault` entirely.

Consequences:

- Existing model-level built-in-search defaults silently revert to off.
- The newly added model also loses its requested default.
- Comments, spacing, ordering, and any representable-but-not-preserved configuration content are rewritten as collateral damage.

## Expected behavior

`add-model` should make a minimal, source-preserving edit: locate the selected provider models list and append the validated model block while retaining unrelated source text.

The resulting complete file should still be parsed and validated before it is atomically written, but validation must not require reconstructing or normalizing the whole document. Configuration-management commands must not silently remove user-authored content outside their declared target.

## Acceptance criteria

- Adding a model preserves existing `webSearchDefault` fields exactly.
- Adding a model preserves comments and unrelated provider/model blocks.
- The operation remains atomic and rejects an invalid resulting configuration without modifying the original file.
- Add focused regression coverage for the above preservation contract.

## Context

This was observed while testing #225. It is independent of Gemini 2.5 Pro native-search/function-tools compatibility.

3aKHP:
Closes #232

## Summary

- replace whole-registry rewrites in every provider/model `vesicle config` mutation with validated, atomic source-preserving edits
- preserve comments, blank lines, ordering, unrelated entries, inline-comment spacing, final-newline semantics, and model fields including `webSearchDefault`; normalize edited output to LF
- keep Guided Setup's existing whole-registry merge-and-normalize behavior, document the boundary in both user languages and current status, and sync bundled `vesicle-docs` references

## Test Plan

- [x] `bun run lint`
- [x] `bun run typecheck`
- [x] `bun test tests/unit/config/provider-source-edit.test.ts tests/integration/cli/config.test.ts` (`58 pass`, `0 fail`)
- [x] `bun test` (`2161 pass`, `10 skip`, `0 fail`)
- [x] `bun run doctor` (`Missing: none`)
- [x] `bun run skills:docs:check` (`85` references in sync)
- [x] `git diff --check`

## Notes / Follow-ups

- This intentionally covers `add-model`, `add-provider`, provider/default `set`, `remove-model`, and `remove-provider`, because all of them shared the same collateral-rewrite and serializer-drift failure class.
- Guided Setup remains the explicit normalization path for onboarding or reconfiguring an endpoint.
- Registry candidates are fully parsed before atomic replacement and reloaded afterward to confirm the round trip.


3aKHP:
## Review follow-up for `ffcbebb20768f06d912a336302b3e08a47cc46cb`

The requested review fixes are now included in the current PR head.

- Removed the obsolete `writeProviderRegistry` path.
- Consolidated candidate validation, optional prerequisite writes, atomic registry replacement, and reload in `commitProviderRegistrySource`; `add-provider` now uses that shared pipeline.
- Removed the leftover positional lookups in the remove commands.
- Made provider-field normalization side-effect-free instead of mutating the loaded registry and serializing it back.
- Reused the canonical YAML unquote helper and a shared provider-block serializer across Setup and source-preserving edits.
- Added exact-source integration coverage for `default.model`, `default.provider`, and provider `defaultModel` synchronization.
- Rejected duplicate top-level sections, default fields, provider fields, and `models:` blocks before any mutation, preventing edits to shadowed values.
- Restored `remove-provider` error precedence so the current-default guard is reported before Semantic Judge references.
- Defined and tested comment ownership when inserting a missing provider field before `models:`: ordinary model documentation remains attached to `models:`, while a commented example for the inserted field remains attached to that field.

The Bot's specific description of where the bundled `userAgent` example was placed did not reproduce against the original source. The underlying ambiguity around the ownership of comments immediately before `models:` was valid, however, and is resolved by the explicit rule and regression coverage above.

Independent CR rechecked only its three original Should-fix findings on this exact head and confirmed all three are resolved. No second Bot Review was requested.

Verification on the current head:

- focused config suites: 104 passed, 0 failed
- full local suite: 2167 passed, 10 skipped, 0 failed
- `bun run lint`, `bun run typecheck`, `bun run doctor`, `bun run skills:docs:check`, and `git diff --check`: passed
- independent CR focused rerun: 64 passed, 0 failed
- GitHub CI: source/package smoke, Linux ELF release smoke, and Windows installer release smoke all passed

PR #248, opened 2026-08-27
test(support): skip host-incapable fixtures by capability probe

3aKHP:
## Summary

- Spreads the 30a84c9 "skip by capability, not platform" doctrine to fixtures that hard-fail on hosts lacking real symlink creation, chmod read denial, POSIX mode bits, absolute-path tar, or /bin/sh.
- Adds shared probes `tests/support/symlink-capability.ts` (`symlinkCapable`) and `tests/support/chmod-capability.ts` (`modeZeroDeniesRead`, `modeBitsPersist`), migrates the two inline probe copies onto them, and adds local probes `tarSupportsAbsoluteArchivePaths` (skills-source) and a `/bin/sh` existence check (pre-turn-auto-compact).
- Splits the composite buffer-io classification and S7 0o600 fixtures so platform-independent assertions keep running on incapable hosts.
- Test-only; no production changes. Every probe succeeds on Linux, so CI coverage there is unchanged. On an unprivileged Windows host the full suite drops from 28 fail to 6 fail (known cross-platform product gaps plus two load-dependent flakes).

## Test Plan

- [x] `bun run lint`
- [x] `bun run typecheck`
- [x] `bun test` — Windows host: 2203 pass / 58 skip / 6 fail, all pre-existing and triaged
- [x] `bun run doctor`

## Notes / Follow-ups

- Production findings triaged while preparing this (deliberately out of scope, each deserves its own PR):
  - `extractTarball` (src/cli/commands/skills/source/archive.ts) passes an absolute archive path to the system tar; GNU tar from Git Bash parses `C:` as a remote host, so GitHub skill installs fail under Git Bash (bsdtar from cmd/PowerShell is fine). A cwd + relative-archive-name fix is portable; `--force-local` would break bsdtar.
  - `path-policy.ts` absolute-path diagnostic takes the "escapes project root" branch for POSIX-style absolutes on Windows (still fail-closed; message differs).
  - `assets.ts` layer shadowing relies on `ENOTDIR`; Windows `lstat` returns `ENOENT` under a regular-file ancestor, so a user-layer file does not mask lower-layer directory reads on Windows.
  - Bun-on-Windows does not enforce cross-process SQLite `BEGIN IMMEDIATE` exclusion (verified with a standalone two-process repro), so the Skill Store index lock is not mutually exclusive across processes on Windows.

### DIFF
--- a/tests/component/tui/workspace/page.test.tsx
+++ b/tests/component/tui/workspace/page.test.tsx
@@ -8,6 +8,7 @@ import { WorkspacePage } from "../../../../src/tui/workspace/view";
 import { createWorkspaceController } from "../../../../src/tui/workspace";
 import { PromptComposer } from "../../../../src/tui/PromptComposer";
 import { paletteFor, reportTerminalThemeMode, setThemePreference } from "../../../../src/tui/theme";
+import { symlinkCapable } from "../../../support/symlink-capability";
 
 let root: string;
 
@@ -205,7 +206,7 @@ describe("tui: workspace page (B2)", () => {
     expect(frame).toContain("inline-renderable");
   });
 
-  test("symlink files show metadata without loading the target", async () => {
+  test.skipIf(!symlinkCapable)("symlink files show metadata without loading the target", async () => {
     await symlink(join(root, "notes.txt"), join(root, "notes-link.txt"));
     const controller = createWorkspaceController(root);
     await controller.openWorkspaceTarget("notes-link.txt");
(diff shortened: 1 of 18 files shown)


3aKHP:
Both nits addressed in b4adaf7: the harness pack test is split so tampered/unlisted/unsafe-manifest assertions run unconditionally (only the symbolic-link assertion stays capability-gated), and the `/bin/sh` gate now uses a functional `Bun.spawnSync(["/bin/sh", "-c", "true"])` probe. Verified locally: both files 0 fail on an unprivileged Windows host, lint/typecheck clean.

PR #285, opened 2026-08-31
fix(agent-loop): background shell completions as provenance-explicit host packets (#284, 1/2)

3aKHP:
## Grade

Standard — core delivery-path + session-record change, dual-track CR per `docs/dev/WORKFLOW.md`.

Refs #284 (delivery-shape half; the idle auto-continuation half follows as a stacked PR on this branch)

## What & why

Background shell completion packets (command + stdout/stderr tails — untrusted process data) persisted as plain `role:"user"` records and rode the wire as bare user text, giving process output user-instruction weight.

- New provenance envelope: host framing sentence (denies user-input semantics) + XML wrapper + escaped per-task blocks carrying task id, originating tool-call id, status, exit code, command, bounded output tail. Record content = wire message body.
- Durable record switches to a `background-process-results` **system** host packet; `history-projector` projects it back into the identical provider-visible user message. All four protocol adapters untouched; legacy user-role records keep projecting unchanged.
- Exactly-once across crash replay: `drainNotifications` split into `collectNotifications` (peek) + `markNotified` (flip only after the record is durable); coverage matched **per task id** (a crash can regrow the replayed batch — only the uncovered complement may append or send; 09345ea).
- `markFailedUserTurn` also fires for a trailing background-results system record, keeping resume/resend free of consecutive user messages (#102 class).

Delivery timing is unchanged in this slice.

## Verification

- [x] `bun run lint`
- [x] `bun run typecheck`
- [x] `bun test` — full suite 2347 pass / 10 skip / 0 fail
- [x] `bun run doctor`
- [x] `bun run skills:docs:sync` (mirror included)

Targeted: `tests/unit/core/agent-loop/background-process.test.ts` (envelope contract), `tests/integration/agent-loop/background-process-notification.test.ts` (system record + identity stamps + crash-replay/regrown-batch/legacy exactly-once), `tests/integration/session/message-history.test.ts` (projection parity + legacy), `tests/integration/agent-loop/shell-permission.test.ts` (wire envelope + record shape), `tests/integration/tui/turn-controller-failed-turn.test.ts` (failed-turn marking).

## Independent CR

- Tier 1 (`vesicle-cr-reviewer`): **0 Blocking**; 1 Should-fix (regrown replay batch double-append via exact-set match) → fixed in 09345ea with a regression test; 1 Nit (unreachable `projectionVisible` trigger named in a comment) → fixed. Verified-claims list covers envelope injection safety (all interpolation escaped), materialize ordering, role-change blast radius, adapter neutrality, and test independence.
- Bot Review (khpilot): left to the owner per the dual-track contract.


3aKHP:
Bot Review nits — addressed in `acfdf18`, except NIT 1 which is answered by the stack rather than changed:

- **NIT 2 (misleading `undelivered` local)** — fixed: renamed to `coverage` (it holds already-delivered state; the real undelivered complement stays in `materialize`).
- **NIT 3 (escaping owned by the scheduler domain)** — fixed: `escapeText`/`escapeAttribute` moved to a neutral owner `src/shared/xml-escape.ts`; both the SubAgent packet renderer and the background-shell envelope import from there, and the scheduler no longer re-exports generic XML helpers.
- **NIT 1 (`exact` looks dead in this PR)** — intentional within the stack: the sole consumer is the idle-delivery path in stacked #286 (`deliverBackgroundProcessResults` uses `persistedCoverage.exact.uuid` as its `prePersistedInputUuid`, so a retried delivery reuses the record instead of appending a second input record — pinned by `background-idle-wake.test.ts` "reuses a pre-persisted background delivery record without appending it twice"). Boundary `materialize` itself never needs it, which is why this PR doesn't read it. The field's doc comment now names that consumer explicitly instead of an unwired promise. Keeping the field here (rather than adding it in #286) keeps the coverage API stable across the two PRs.

Full suite re-verified on both branches after the refactor (`acfdf18` / #286 rebased): lint, typecheck, and 2356 tests / 0 fail.

